PR #2705, opened 2025-05-04
:hammer: Add koin-ktor to improve router implementation and avoid bloated parameter passing

### LINKED ISSUE #2704: Add koin-ktor to improve router implementation and avoid bloated parameter passing

guiyanakuang:
close #2704


guiyanakuang:
koin-ktor is not support in iOS

PR #2728, opened 2025-05-14
Add Persian Language

amir1376:
Hi! A big thank you for your great project 👍

I couldn't find any info about how can I contribute translations. ( I was looking maybe there is a third party translation platform like [Crowdin](https://crowdin.com) you are using. but I didn't find anything, that's why created this PR here )

in this PR I've added Persian (fa) Language. if it requires any adjustments just let me know.


guiyanakuang:
@amir1376. Thanks for the Persian translation PR. Merged.

Note that CrossPaste UI wasn't initially designed with RTL support in mind. We'll look into adding proper RTL handling in a future release to better support Persian and other RTL languages.

Appreciate the contribution.

PR #2961, opened 2025-07-12
:sparkles: Add option to change application font

amir1376:
This pull request introduces a setting that allows users to change the font used in the application interface

### DIFF
--- a/app/src/commonMain/kotlin/com/crosspaste/ui/settings/FontSettingItemView.kt
+++ b/app/src/commonMain/kotlin/com/crosspaste/ui/settings/FontSettingItemView.kt
@@ -153,8 +153,8 @@ fun FontSettingItemView() {
                             .heightIn(max = 300.dp)
                             .wrapContentWidth()
                             .background(Color.Transparent)
-                            .clip(tiny2XRoundedCornerShape)
-                            .shadow(small),
+                            .shadow(small)
+                            .clip(tiny2XRoundedCornerShape),
                 ) {
                     val maxWidth =
                         max(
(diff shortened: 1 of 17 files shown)


amir1376:
Currently, when I want to add a new string, I have to provide translations for all supported languages—even the ones I don’t speak—because tests fail if any key is missing.
I think it would be better to allow missing keys to fall back to the default (en.properties). This avoids using low-quality auto-translations just to pass tests, and lets native speakers contribute proper translations over time.

guiyanakuang:
@amir1376. Thanks for bringing this up! I completely agree with your point about translation fallbacks. Requiring translations for all languages upfront does create unnecessary friction for contributors, and auto-translated strings are definitely not ideal.

I'll work on implementing automatic fallback to en.properties for missing keys. This approach makes much more sense - it allows us to ship features faster while leaving room for native speakers to contribute quality translations when they can.

Thanks again for this contribution! The font customization feature is a great addition.

app/src/commonMain/kotlin/com/crosspaste/ui/settings/FontSettingItemView.kt
```diff
@@ -0,0 +1,218 @@
...
+                val isScrolling = scrollState.isScrollInProgress
+                Box(
+                    modifier =
+                        Modifier
+                            .wrapContentHeight()
+                            .heightIn(max = 300.dp)
+                            .wrapContentWidth()
+                            .background(Color.Transparent)
+                            .clip(tiny2XRoundedCornerShape)
+                            .shadow(small),
```
guiyanakuang, line 157: I noticed something while testing - it seems like the shadow might not be showing up as expected. I think this could be related to the modifier order here

Maybe we could try swapping them

.shadow(small)
.clip(tiny2XRoundedCornerShape)

amir1376: Fixed.

guiyanakuang:
Merged! 🎉
Thank you so much for this contribution and for the quick fix on the shadow issue. The font customization feature is a wonderful addition that I'm sure many users will appreciate.

CompileFuture2024:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
You're welcome! Glad I could help

PR #2965, opened 2025-07-12
:hammer: Move JVM-bound implementation from commonMain to DesktopMain

### LINKED ISSUE #2964: Move JVM-bound implementation from commonMain to DesktopMain

guiyanakuang:
- Ensure commonMain remains cross-platform compatible
- Support compilation for Android and iOS on mobile platforms

close #2964


app/src/commonMain/kotlin/com/crosspaste/ui/base/FontManager.kt
```diff
@@ -98,18 +57,16 @@ abstract class FontManager(
             ?: defaultFontInfo
 
     fun setFont(fontId: String?) {
```
guiyanakuang, line 59: setFont is only called by UI single thread, so no concurrency control needed

PR #4302, opened 2026-04-26
:bug: use imageFlavor instead of javaFileListFlavor for single-image clipboard writes

### LINKED ISSUE #2729: 截图同步到另一个mac再拷贝不能粘贴到微信聊天输入框
### How frequently does the bug occur?

Always

### Description

截图同步到另一个mac再拷贝不能粘贴到微信聊天输入框

### Stacktrace & log output

```shell

```

### Can you reproduce the bug?

Always

### Reproduction Steps

前提
1、两台mac电脑A和B连接同一个wifi网络，其中一台需要安装微信
复现步骤
1、在A上面截图，任何截图工具都可以，截图后数据在内存中的
2、打开crosspaste，检查是否已经同步到B上面
3、已经同步的话，就单击拷贝或者右键再点击拷贝
4、然后打开微信，command+v
5、问题出现，并不能粘贴到微信聊天输入框中

但是我在crosspaste中打开文件然后复制粘贴文件到微信也是可以的，但是这样就多了一个步骤很麻烦
就没有使用这个app的必要了

直接在电脑B上截图再发到微信是可以的

### Version

crosspaste 1.1.2(1357)

### OS

macos 15.5

Fldicoahkiin:
## Summary

Fixes #2729

When pasting a single image from CrossPaste history (right-click copy or double-click quick paste), apps like WeChat and Notes on macOS receive a `file:///` path instead of the actual image. This is because `DesktopImageTypePlugin.buildTransferable()` places `javaFileListFlavor` (file list) on the clipboard, which macOS translates to `public.file-url` pasteboard entries. Apps that prefer text/URL types over image types end up pasting the file path.

This PR changes single-image clipboard writes to use `imageFlavor` (BufferedImage) as the primary clipboard data, matching the behavior of native macOS screenshot paste (`public.tiff`). Falls back to `javaFileListFlavor` only when the image cannot be read. Multi-image clipboard writes remain unchanged.

## Changes

- For single images: place `imageFlavor` (BufferedImage) on the clipboard instead of `javaFileListFlavor`
- Fall back to `javaFileListFlavor` when `imageHandler.readImage()` returns null or throws
- Multi-image behavior unchanged (still uses `javaFileListFlavor`)

## Test plan

- [x] Take a screenshot on macOS, copy from CrossPaste history, paste into WeChat — image pastes correctly
- [x] Take a screenshot on macOS, copy from CrossPaste history, paste into Notes — image pastes correctly
- [ ] Verify multi-image paste still works as file list
- [ ] Verify drag-and-drop from side panel still works
- [ ] Verify on Windows and Linux that existing behavior is preserved

### DIFF
--- a/app/src/desktopMain/kotlin/com/crosspaste/paste/plugin/type/DesktopImageTypePlugin.kt
+++ b/app/src/desktopMain/kotlin/com/crosspaste/paste/plugin/type/DesktopImageTypePlugin.kt
@@ -219,25 +219,20 @@ class DesktopImageTypePlugin(
         val fileList: List<File> = filePaths.map { it.toFile() }
 
         if (fileList.size == 1) {
-            var imageAdded = false
             runCatching {
                 val start = System.currentTimeMillis()
                 val image: BufferedImage? = imageHandler.readImage(fileList[0].toOkioPath())
                 image?.let {
                     map[DataFlavor.imageFlavor.toPasteDataFlavor()] = it
-                    imageAdded = true
                 }
                 val end = System.currentTimeMillis()
                 logger.debug { "read image ${fileList[0].absolutePath} use time: ${end - start} ms" }
             }.onFailure { e ->
                 logger.error(e) { "read image fail" }
             }
-            if (!imageAdded) {
-                map[DataFlavor.javaFileListFlavor.toPasteDataFlavor()] = fileList
-            }
-        } else {
-            map[DataFlavor.javaFileListFlavor.toPasteDataFlavor()] = fileList
         }
+
+        map[DataFlavor.javaFileListFlavor.toPasteDataFlavor()] = fileList
 
         if (mixedCategory) {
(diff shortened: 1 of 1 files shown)


guiyanakuang:
Hi @Fldicoahkiin, thanks a lot for the PR and the careful diagnosis — #2729 is a real bug, and you've correctly identified that AWT translates javaFileListFlavor into public.file-url, which apps like WeChat and Notes prefer over the image flavor. Aiming for parity with native screenshot paste is a sensible direction.  

I pulled the branch locally to test and unfortunately found a regression we need to address before merging:                                                                                                                                                                            
  - Pasting a single image into Finder no longer creates a file. Finder relies on javaFileListFlavor to materialize the paste. With the current PR, single-image clipboards only carry imageFlavor, so Finder has nothing to work with.  
  - Drag-and-drop from CrossPaste's side panel onto Finder uses the same buildTransferable and is affected the same way.  

So we still need to keep `map[DataFlavor.javaFileListFlavor.toPasteDataFlavor()] = fileList` in the single-image branch.

PR #4466, opened 2026-05-23
:bug: push protocol: remove pull fallback + gate behind forcePush opt-in

guiyanakuang:
## Summary

Follow-up fixes to two bugs in the M4 push protocol surfaced during iOS Share Extension real-device testing. Both fixes are in `commonMain` so they sync to mobile via `./gradlew syncDesktop`.

Background design doc: `crosspaste-mobile/docs/push-protocol-issues.md` — full root-cause analysis, the `forcePush` field design, and the migration / compat story. This PR covers **problems 1 and 2** from that doc; problem 3 (mobile-only) is tracked separately.

### Bug 1 — push failure no longer falls back to pull-mode metadata send

`SyncPasteTaskExecutor.syncPasteToTarget` previously fell through to `pasteClientApi.sendPaste` when `filePushService.pushFiles` returned a `FailureResult`. For senders that cannot host reverse pulls (notably iOS Share Extension) this created a doomed second LOADING entry on the receiver plus a reverse `/pull/file` that always failed. Same-version peers always implement the M1+M2 server endpoints, so the fallback never helped. Push failure now returns the `FailureResult` unchanged and the SyncTask retries on the next executor pass.

### Bug 2 — push path is now opt-in via `SyncExtraInfo.forcePush`

Previously **any** file-type paste between same-version peers was forced through push, even for callers that can host a PasteServer (main iOS app, Android, desktop-to-desktop). Push has a longer failure surface (prepare → multi-chunk → complete) and is only required when the sender cannot host reverse pulls. Push is now opt-in: only callers that explicitly set `forcePush = true` take the push path.

Default behavior for all existing call sites is unchanged (pull-mode metadata send). The only legitimate caller for `forcePush = true` is iOS Share Extension, which will set it in the follow-up mobile PR.

## Wiring (all commonMain)

| Class / field | Change |
|---|---|
| `SyncExtraInfo` | new `val forcePush: Boolean = false` |
| `TaskBuilder.addSyncTask` | new `forcePush: Boolean = false` param |
| `DesktopTaskBuilder.addSyncTask` | implements new param, forwards into `SyncExtraInfo` |
| `PasteSourceContext` | new `val forcePush: Boolean = false` |
| `PasteReleaseService.releaseLocalPasteData` | signature widened to take the full `PasteSourceContext` (cleaner than threading two related fields); forwards `forcePush` into `addSyncTask` |
| `PasteCollector.completeCollect` | passes the captured `sourceContext` instead of just `targetAppInstanceIds` |
| `SyncPasteTaskExecutor.shouldUsePushPath` | now takes `SyncExtraInfo`; returns `syncExtraInfo.forcePush && pasteData.isFileType() && handler.currentVersionRelation == EQUAL_TO` |
| `SyncPasteTaskExecutor.syncPasteToTarget` | push branch no longer falls through to `sendPaste` on failure |

All existing desktop call sites continue to use the default `forcePush = false`.

## Backward compatibility

- Old DB-serialized `SyncExtraInfo` rows that don't have a `forcePush` field deserialize to `false` (the kotlinx-serialization default) → pull-mode behavior preserved.
- Old mobile clients that don't know about `forcePush` send tasks via pull mode → still works.
- No wire-format changes to `PushPrepareResponse`, `PushHeaders`, or other push protocol artifacts.

## Test plan

- [x] `./gradlew ktlintFormat` — clean
- [x] `./gradlew :app:compileKotlinDesktop :app:compileTestKotlinDesktop` — clean
- [x] `./gradlew :app:desktopTest --tests "com.crosspaste.task.SyncPasteTaskExecutorTest"` — 18/18 pass, including the 4 new forcePush cases:
  - `doExecuteTask_forcePushTrue_fileTypeEqualToNonExtension_takesPushPath`
  - `doExecuteTask_forcePushTrue_pushFails_returnsFailureNoFallback`
  - `doExecuteTask_forcePushFalse_fileTypeEqualTo_routesToSendPaste`
  - `doExecuteTask_forcePushTrue_nonFileType_routesToSendPaste`
- [x] New `DesktopTaskBuilderTest` (3 tests, all pass) verifies forcePush propagates from `TaskBuilder.addSyncTask` into the queued `SyncExtraInfo`.
- [x] `PasteCollectorTest` extended with a case capturing the forwarded `PasteSourceContext` and asserting `forcePush` flows through.
- [x] Wider regression: `./gradlew :app:desktopTest --tests "com.crosspaste.task.*" --tests "com.crosspaste.sync.*" --tests "com.crosspaste.net.*" --tests "com.crosspaste.paste.*"` — 539 tests, 0 failures, 0 errors, 15 skipped (pre-existing).

## Deviations from spec

- The spec mentions a `MobileTaskBuilder` implementation; none exists in this repo (mobile lives in a separate private repo and consumes synced commonMain). Only `DesktopTaskBuilder` was updated here.
- Replaced the existing `doExecuteTask_pushFails_fallsBackToPullSend` test (which now describes the removed behavior) rather than just renaming it — its assertion would be inverted otherwise.
- The pre-existing `doExecuteTask_fileTypeEqualToNonExtension_takesPushPath` case was renamed to `doExecuteTask_forcePushTrue_fileTypeEqualToNonExtension_takesPushPath` and now passes `forcePush = true` explicitly.

## Follow-up

- Mobile-side problem 3 (iOS Share Extension bypassing `SyncManager` / `SyncTask`) is tracked separately and will land in `crosspaste-mobile` after this PR syncs.

Refs: PR #4457 (M4 mobile push client), PR #4456 (M1+M2 server push).

### DIFF
--- a/app/src/commonMain/kotlin/com/crosspaste/paste/PasteCollector.kt
+++ b/app/src/commonMain/kotlin/com/crosspaste/paste/PasteCollector.kt
@@ -118,7 +118,7 @@ class PasteCollector(
                                 }
                             }
                     }
-                    pasteReleaseService.releaseLocalPasteData(id, pasteItems, sourceContext.targetAppInstanceIds)
+                    pasteReleaseService.releaseLocalPasteData(id, pasteItems, sourceContext)
                 }.onFailure { e ->
                     logger.error(e) { "Failed to complete paste $id" }
                     markDeletePasteData(id)
--- a/app/src/commonMain/kotlin/com/crosspaste/paste/PasteReleaseService.kt
+++ b/app/src/commonMain/kotlin/com/crosspaste/paste/PasteReleaseService.kt
@@ -76,8 +76,9 @@ class PasteReleaseService(
     suspend fun releaseLocalPasteData(
         id: Long,
         pasteItems: List<PasteItem>,
-        targetAppInstanceIds: Set<String>?,
+        sourceContext: PasteSourceContext,
     ) = withContext(ioDispatcher) {
+        val targetAppInstanceIds = sourceContext.targetAppInstanceIds
         pasteDao.getLoadingPasteData(id)?.let { pasteData ->
             var pasteAppearItems = pasteItems
             for (pastePlugin in pasteProcessPlugins) {
@@ -135,7 +136,13 @@ class PasteReleaseService(
                     // items skip Apple peers).
                     val hasValidSyncTargets = targetAppInstanceIds == null || targetAppInstanceIds.isNotEmpty()
                     if (!pasteData.remote && hasValidSyncTargets) {
-                        addSyncTask(id, maxFileSize, pasteData.appInstanceId, targetAppInstanceIds)
+                        addSyncTask(
+                            id,
+                            maxFileSize,
+                            pasteData.appInstanceId,
+                            targetAppInstanceIds,
+                            forcePush = sourceContext.forcePush,
+                        )
                     }
 
                     if (pasteType.isFile() || pasteType.isImage()) {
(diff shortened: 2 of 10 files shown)


guiyanakuang:
Superseded by #4467: the forcePush opt-in design is dropped in favor of pulling push out of `SyncPasteTaskExecutor` entirely into a standalone `SharePushOrchestrator` service. See #4467 for the full rationale (push and pull have fundamentally different sync models).

Issue #4509, opened 2026-06-05
Network discovery (mDNS) does not recover after the network becomes available or the IP changes


guiyanakuang:
## Summary

When the desktop app starts **before the network is ready**, or when the device's **IP changes while running** (e.g. macOS *Private Wi-Fi Address* rotation, DHCP lease renewal, Wi-Fi roaming), local discovery (mDNS) silently stays down for the entire session. The app keeps probing the *stale* stored peer IP and never rediscovers the peer, so devices remain offline until the app is restarted while the network is up.

Observed in #4499 (logs from both machines):

- Both devices got new DHCP IPs (Mac `…52`→`…84`, Windows `…64`→`…62`); device identities never changed.
- On the Mac, after a cold start with no network: `0` local interfaces detected, `0` mDNS services resolved, DNS `UnresolvedAddressException` at startup — for the **whole session** — while it logged `no reachable host` 700+ times against the stale IP.
- Deleting + restarting the app fixed it (by then the network was up).

## Root cause

The mDNS service already rebuilds itself whenever the interface list changes — `DesktopPasteBonjourService.init` collects `networkInterfaceService.networkInterfaces` and does `close() + setup()` on each emission. The problem is that the flow feeding it **does not reflect real OS network state**:

1. `DesktopNetworkInterfaceService.networkInterfaces` is derived **only** from config:
   ```kotlin
   configManager.config.map { it.useNetworkInterfaces }.distinctUntilChanged()
       .mapLatest { createNetworkInterfaceInfos(it) }
   ```
   It re-emits only when the config field `useNetworkInterfaces` changes. OS-level interface up/down, IP changes, and NIC add/remove are invisible to it.
2. Startup empty value gets locked in: `initNetworkInterfaceInfos()` returns an empty list when `getAllNetworkInterfaceInfo()` is empty (network not ready) and skips auto-select. Once the network comes up, the config hasn't changed, so the flow never re-emits.
3. There is **no connectivity/interface-change observer** and no periodic refresh. `clearProviderCache()` is never called in production code.
4. Consequence: mDNS stays down → no fresh peer addresses → `SyncResolver.discoverReachableHost()` keeps telnet-probing the stale stored `hostInfoList` → permanent `no reachable host`.

## Why "wait for network at startup" alone is not enough

[...]

PR #4519, opened 2026-06-05
:sparkles: push subnet-matched address on /sync/telnet to cut mDNS reconnection latency

### LINKED ISSUE #4518: Push subnet-matched address on /sync/telnet to cut mDNS reconnection latency
## Background

Phase 3 of the network discovery work (relates to #4509). mDNS multicast has latency and loss, so after a device changes IP a peer can wait up to the next mDNS round before it learns the new address. Most IP changes are single-sided: the device that moved still knows where its peer is and will telnet it to reconnect — so it can piggyback its own current address on that probe and let the peer update immediately.

This is a latency optimization layered on top of discovery self-healing — it does **not** change any protocol version and is fully backward compatible (older peers ignore the extra header).

## Change

1. **Client** (`TelnetHelper`): every `/sync/telnet` probe carries a `crosspaste-sync-info` header advertising our **subnet-matched** address — the local interface(s) on the peer's subnet, selected via the existing `HostInfo.filter(peerAddress)` (the same subnet match the trust flow already uses). The port comes from our own `SyncInfo`, so the peer learns IP **and** port. When no interface shares the peer's subnet (cross-subnet / routed), nothing is advertised. Building the hint is best-effort and time-bounded so it can never delay or fail the probe (notably during the cold-start window before our own server port is ready).
2. **Server** (`SyncRouting` `/sync/telnet`): decodes the header and merges the advertised address through the **same `NearbyDeviceManager.addDevice` path mDNS uses**. This keeps the trust model unchanged: it is an unauthenticated routing hint (identical trust level to an mDNS TXT record), so an unknown peer only populates the in-memory nearby list (never the DB) and an already-known peer gets its `hostInfoList` merged. Trust is still granted solely by the ECDH heartbeat.
3. A shared `SyncInfoHeaderCodec` owns the wire format (Base64-wrapped JSON) used by both sides.

Note: the **source-IP harvesting** originally sketched as a separate step is intentionally dropped — this subnet-matched push supersedes it (it yields IP *and* port authoritatively, where source-IP yields only an IP with the port inferred).

## Tests

- `TelnetHelperTest`: advertises only the subnet-matched interface; omits the header when no interface shares the peer's subnet.
- `SyncInfoHeaderCodecTest`: encode/decode round-trip and garbage-tolerance.
- Existing server-merge behavior is covered by `GeneralNearbyDeviceManager` `addDevice` tests; `SyncIntegrationTest` exercises the real telnet route end-to-end.
- Full `app:desktopTest` suite green; ktlint clean.

guiyanakuang:
Closes #4518

Phase 3 of the network discovery work (relates to #4509). mDNS multicast has latency and loss, so after a device changes IP a peer can wait up to the next mDNS round before learning the new address. Most IP changes are **single-sided**: the device that moved still knows where its peer is and telnets it to reconnect — so it piggybacks its current address on that probe and the peer updates immediately.

A latency optimization on top of discovery self-healing — **no protocol version change**, fully backward compatible (older peers ignore the extra header).

## Changes

1. **Client** (`TelnetHelper`): every `/sync/telnet` probe carries a `crosspaste-sync-info` header advertising our **subnet-matched** address — the local interface(s) on the peer's subnet, selected via the existing `HostInfo.filter(peerAddress)` (the same subnet match the trust flow already uses). The port rides along from our `SyncInfo`, so the peer learns IP **and** port. Nothing is advertised when no interface shares the peer's subnet (cross-subnet / routed). Building the hint is best-effort and bounded by `ADVERTISE_BUILD_TIMEOUT`, so it can never delay or fail the probe — notably during the cold-start window where `createEndpointInfo` would otherwise block on `portFlow` until our own server is up.
2. **Server** (`SyncRouting` `/sync/telnet`): decodes the header and merges through the **same `NearbyDeviceManager.addDevice` path mDNS uses**. Trust model unchanged — an unauthenticated routing hint (same level as an mDNS TXT record): unknown peers only populate the in-memory nearby list (never the DB); already-known peers get their `hostInfoList` merged. Trust is still granted solely by the ECDH heartbeat.
3. **`SyncInfoHeaderCodec`**: shared `Base64(JSON)` wire codec used by both sides; `clientSyncInfo()` now decodes through it.

Source-IP harvesting (originally sketched as a separate step) is intentionally dropped — this subnet-matched push supersedes it (IP **and** port, authoritatively).

## Tests

- `TelnetHelperTest`: advertises only the subnet-matched interface; omits the header when no interface shares the peer's subnet.
- `SyncInfoHeaderCodecTest`: encode/decode round-trip + garbage tolerance.
- Server-merge behavior covered by existing `GeneralNearbyDeviceManager` `addDevice` tests; `SyncIntegrationTest` exercises the real telnet route.
- Full `app:desktopTest` green; ktlint clean.

> Independent of the still-open #4510 (phase 1/5) and #4515 (phase 2) — touched files are disjoint, so this can be reviewed/merged on its own.

### DIFF
--- a/app/src/commonMain/kotlin/com/crosspaste/net/DefaultServerModule.kt
+++ b/app/src/commonMain/kotlin/com/crosspaste/net/DefaultServerModule.kt
@@ -107,6 +107,7 @@ open class DefaultServerModule(
                     appTokenApi,
                     configManager,
                     exceptionHandler,
+                    nearbyDeviceManager,
                     networkInterfaceService,
                     pendingKeyExchangeStore,
                     secureKeyPairSerializer,
--- a/app/src/commonMain/kotlin/com/crosspaste/net/SyncInfoHeaderCodec.kt
+++ b/app/src/commonMain/kotlin/com/crosspaste/net/SyncInfoHeaderCodec.kt
@@ -1,0 +1,36 @@
+package com.crosspaste.net
+
+import com.crosspaste.dto.sync.SyncInfo
+import com.crosspaste.utils.getJsonUtils
+import kotlinx.serialization.decodeFromString
+import kotlinx.serialization.encodeToString
+import kotlin.io.encoding.Base64
+import kotlin.io.encoding.ExperimentalEncodingApi
+
+/**
+ * Wire codec for advertising a [SyncInfo] inside an HTTP header. The payload is the
+ * JSON encoding of the SyncInfo, Base64-wrapped so it survives as a single header value.
+ *
+ * Two sites must agree byte-for-byte on this format:
+ *  - clients attach [HEADER] on `/sync/telnet` to push their current (subnet-matched)
+ *    address so a peer learns it without waiting for the next mDNS round (#4509 phase 3),
+ *  - the server decodes it back in `SyncRouting` (also reused by the trust handshake).
+ *
+ * Keeping encode/decode in one object is what guarantees that symmetry.
+ */
+object SyncInfoHeaderCodec {
+
+    const val HEADER: String = "crosspaste-sync-info"
+
+    @OptIn(ExperimentalEncodingApi::class)
+    fun encode(syncInfo: SyncInfo): String =
+        Base64.encode(getJsonUtils().JSON.encodeToString(syncInfo).encodeToByteArray())
+
+    @OptIn(ExperimentalEncodingApi::class)
+    fun decode(encoded: String): SyncInfo? =
+        runCatching {
+            getJsonUtils().JSON.decodeFromString<SyncInfo>(
+                Base64.decode(encoded).decodeToString(),
+            )
+        }.getOrNull()
+}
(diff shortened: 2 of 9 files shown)


guiyanakuang:
Follow-up (commit fa25a276) addressing the review concern that every header-bearing probe re-runs server-side \`addDevice\` → \`trackSignificantAction\` (no-op on desktop, but copied to mobile where telnet's higher frequency than mDNS could inflate the significant-action count toward a premature rating prompt).

Fixed at the source rather than relying on the mobile impl: \`TelnetHelper\` now advertises to each peer **once** and stays quiet until our address actually changes. It caches, per probed peer address, the subnet-matched host list it last delivered (on a 200), and attaches the header only when the freshly computed list differs — a steady network advertises exactly once; a real IP change recomputes a different list and re-advertises on the next probe. The subnet match is split into a cheap non-blocking selector (drives the change check) from the port-bound `SyncInfo` encode (only when we actually send).

New tests: no re-advertise when unchanged; re-advertise after the local address changes.

PR #4539, opened 2026-06-09
:sparkles: Ship a root-level portable shortcut in the Windows zip

### LINKED ISSUE #4538: Windows portable zip: provide a root-level launcher shortcut
## Problem
The Conveyor-built Windows portable zip places the launcher at
`bin\CrossPaste.exe`. After extracting the zip, users have to navigate into
`bin\` to find and run the executable, which is unintuitive next to the
`bin/` / `app/` folders.

## Proposal
Ship a portable `CrossPaste.lnk` at the zip root that points to
`bin\CrossPaste.exe` via a **relative** path, so a freshly extracted zip has an
obvious launch entry at the top level regardless of where it is extracted.

## Notes
- The `.lnk` carries a sentinel absolute target only so Explorer treats it as a
  valid shortcut; when that path is absent it falls back to the relative path,
  resolved against the `.lnk`'s own directory (the zip root).
- The shortcut is version-independent, generated once and committed; a
  reproducible generator (`scripts/generate-windows-lnk.py`, pylnk3) is included
  for regeneration.
- Relative-fallback launch behaviour still needs validation on real Windows
  (Win10/Win11, non-C: drive, after moving the folder).

guiyanakuang:
Closes #4538

## What
The Conveyor-built Windows portable zip places the launcher at `bin\CrossPaste.exe`, so a freshly extracted zip forces users to dig into `bin\` to start the app. This ships a portable `CrossPaste.lnk` at the **zip root** that points to `bin\CrossPaste.exe` via a **relative** path, giving an obvious launch entry next to `bin/` / `app/`.

## How it resolves on Windows
- The `.lnk` carries a sentinel absolute target (`C:\CrossPaste\bin\CrossPaste.exe`) only so Explorer treats it as a valid, double-clickable shortcut.
- It also carries a relative path (`.\bin\CrossPaste.exe`). When the absolute target is missing, the shell falls back to the relative path resolved against the directory containing the `.lnk` (the zip root), so it works no matter where the user extracts or moves the folder.

## Implementation notes
- The shortcut is **version-independent** (the relative target never changes), so it is generated once and committed at `app/script/CrossPaste.lnk`; `conveyor.conf` maps it to the zip root.
- `scripts/generate-windows-lnk.py` (pylnk3) is a reproducible generator — timestamps are pinned so the committed binary is byte-for-byte stable and leaks no build-host info. It is only needed to regenerate, **not** in CI.
- `work_dir` / icon are intentionally omitted (relative `work_dir` would point cwd at a non-existent path; with no icon set, Explorer uses the resolved target exe's own icon).

## Validation needed (manual, on real Windows)
The `.lnk` is generated cross-platform, but the relative-fallback **launch** behaviour must be verified on Windows before relying on it:
- [ ] Extract under `C:\`, double-click the root `CrossPaste.lnk` → launches `bin\CrossPaste.exe`
- [ ] Extract on a **non-C: drive** (e.g. `D:\tools\`) → still launches (real test of relative fallback)
- [ ] **Move** the extracted folder, then double-click → still launches
- [ ] Win10 and Win11
- [ ] No "searching for this item" dialog / noticeable delay on double-click
- [ ] Icon renders acceptably (generic icon is acceptable)


guiyanakuang:
Abandoning this approach. A static prebuilt .lnk cannot portably point to bin\CrossPaste.exe: on Windows the absolute IDList target is what actually launches, and the relative-path fallback proved unreliable ("Windows can't find the target"), with no icon since the icon is taken from the unresolved target. A position-independent launch requires either a %~dp0-based launcher or runtime shortcut creation by the app — out of scope for now.

PR #4654, opened 2026-07-23
:sparkles: Pairing v3 Phase 3: network protocol service, routing, and client API

### LINKED ISSUE #4653: Pairing v3 Phase 3: network protocol service, routing, and client API
## Summary

Phase 3 of the pairing v3 protocol (design: \`ai/docs/PAIRING_PROTOCOL_V3.md\` §19, foundation shipped in #4651 / #4652): wire the dormant Phase 1/2 building blocks into an actual network protocol.

## Scope

- **PairingProtocolV3Service** (app commonMain): orchestrates both roles
  - Acceptor: handle intent (validation, rate limit, acceptance window, session + PIN + PAKE share + signed offer creation), handle proof (PAKE derive, transcript recompute, key confirmation, identity signature), handle commit (idempotent receipt, key persistence), advisory cancel
  - Initiator: send intent, validate offer, submit PIN (PAKE + proof + verify acceptor proof + commit + verify receipt + persist key), offer refresh after PIN rotation
  - PIN rotation driver (30s generations, atomic publication via \`rotateGeneration\`) and session expiry sweep
- **PairingV3Routing**: \`POST /sync/pairing/v3/{intent,proof,commit,cancel}\` — thin handlers delegating to the service
- **PairingV3ClientApi**: transport-only client following the existing \`request()\`/\`ClientApiResult\` conventions
- **Strict DTO validation and error mapping**: v3 error codes added to \`StandardErrorCode\`; proof-stage failures collapse to \`PAIRING_PROOF_INVALID\` on the wire (design §15)
- **No-downgrade guard** (design §17.2): while a v3 acceptor session is active for a peer, v2 trust endpoints reject that peer
- **Koin wiring**: session store, rate limiter, receipt cache, acceptance window, service, client API

## Explicitly out of scope

- Real SPAKE2 \`PakeProvider\` (tracked separately; production wiring uses a throwing placeholder, the v3 surface stays inert: capability advertisement remains \`PAIRING_VERSION = 2\` and the acceptance window defaults to closed)
- Desktop UI (Phase 4), mobile integration (Phase 5), rollout flag (Phase 6)

## Test plan

- Two-instance real-server integration tests (existing \`TestInstance\` harness) with the deterministic fake PAKE: happy path, wrong PIN + generation invalidation + recovery, rotation, duplicate/replayed intents, capacity, rate limiting, idempotent commit retries, downgrade guard
- Service-level validation matrix for malformed intents/proofs/commits

guiyanakuang:
Closes #4653

Phase 3 of the pairing v3 protocol (design: `ai/docs/PAIRING_PROTOCOL_V3.md` §19; foundation: #4652). Wires the dormant Phase 1/2 building blocks into an actual network protocol. **The v3 surface stays inert in production**: capability advertisement remains `PAIRING_VERSION = 2`, the acceptance window defaults to closed, and the production Koin wiring uses a throwing `UnavailablePakeProvider` until the reviewed SPAKE2 provider lands.

### `PairingProtocolV3Service` (app commonMain)
Orchestrates both roles; routing and the client API are pure transport.

- **Acceptor**: intent validation (shape, target/caller binding, signature, ciphersuite) → rate limit → session + PIN + PAKE share + signed offer creation; proof handling (PAKE derive, transcript recomputed locally — peer-supplied hashes are never adopted, key confirmation, identity signature); commit with idempotent receipts; advisory cancel.
- **Initiator**: `startPairing` (signed intent → offer validation incl. signature and `requestHash` echo), `submitPin` (PAKE → proof → verify acceptor proof → commit → verify receipt → persist key), `refreshOffer` (byte-identical intent resend adopts the current generation after rotation), `retryCommit` (commit bytes are deterministic; interrupted commits resume without a new PIN).
- **Drivers**: per-session 30s PIN rotation loop (atomic generation publication via `rotateGeneration`, early wake-up nudge after a failed proof) and a lazy expiry/prune maintenance sweep.

### Protocol behaviors
- Byte-identical intent retry is idempotent and returns the current generation's exact signed offer (this is also how the initiator picks up a rotated PIN).
- A fresh intent from the same verified signing key cancels and replaces the stale session (signature-gated restart UX).
- Commit receipts are cached, **bound to the committing peer's app instance id**, and byte-identical retries return the original authenticated receipt — including after the terminal session card was pruned; different bytes or a different caller never resolve.
- Proof-stage failures collapse to `PAIRING_PROOF_INVALID` on the wire (design §15); one failed proof invalidates the generation and immediately rotates a fresh PIN.
- No-downgrade guard (§17.2): while a v3 acceptor session is active for a peer, `/sync/trust`, `/sync/trust/v2/exchange`, and `/sync/trust/v2/confirm` refuse that peer with `PAIRING_VERSION_UNSUPPORTED`.

### Transport
- `PairingV3Routing`: `POST /sync/pairing/v3/{intent,proof,commit,cancel}`; successful commit triggers the same `trustSyncInfo` path as v2.
- `PairingV3ClientApi`: transport-only, following the existing `request()`/`ClientApiResult` conventions.
- 15 v3 error codes added to `StandardErrorCode` (3200 range) with a 1:1 locked mapping to `PairingV3ErrorCode`; new `PairingCancelV3` DTO (advisory cancel; an authenticated cancel MAC would need a new frozen key-schedule label and stays an explicit open decision).

## Independent review
A strict review pass found 1 P1 + 3 P2, all fixed in the second commit:
- **P1**: pruned-session commit path resolved receipts by `(sessionId, commitMac)` alone — receipts are now peer-bound end to end.
- **P2**: downgrade guard extended to `/sync/trust/v2/confirm`; maintenance-loop stop/start race fixed (mutex-guarded re-check); rotation now publishes store generation + current offer atomically w.r.t. intent-retry reads.

Deliberately deferred (noted for follow-up): address-level pre-rate-limit before intent signature verification; commit transport-interruption e2e simulation.

## Tests
- 16 two-instance real-server integration tests (`PairingV3IntegrationTest`, deterministic fake PAKE): happy path (+ encrypted heartbeat after pairing), window gate, validation matrix, wrong PIN → generation invalidation → rotation → recovery, concurrency/capacity/restart-replace, rotation + stale-proof refusal, cancel propagation, byte-level commit idempotency, terminal intent replay, tampered proof + attempt budget, rate limiting, peer-bound pruned-receipt replay, v2 exchange/confirm downgrade guard.
- Receipt-cache peer-binding unit tests; error-mapping lock test; `PairingCancelV3` serialization round-trip.
- Full `app:desktopTest`, `core:desktopTest`, `core:nativeAppTest`, and `ktlintCheck` pass.


guiyanakuang:
Third commit (`d134ef6a`) — concurrency hardening after another independent review pass:

- **Per-peer version coordination lock** (`PairingVersionCoordinator`, striped mutex shared by `syncRouting` and `pairingV3Routing`): v1 trust, v2 exchange/confirm, and v3 intent for the same peer are now serialized, closing interleavings where legacy trust could complete concurrently with v3 session creation. The downgrade guard now covers both roles (`hasActiveSession`), a successful v3 intent clears any stale pending v2 exchange, and plain v2 pairing without any v3 session remains compatible (covered by a dedicated test).
- **Trust finalization is atomic**: `PairingSessionStore.transition` gained a mutex-guarded `guard` hook; commit-side key persistence, receipt registration, and COMMITTING → TRUSTED now run inside the session mutex and the final transition result is checked. A reject that wins the race is honored (`PAIRING_REJECTED`, no key persisted, no ack); a trust that entered finalization first makes the reject observe TRUSTED. Persistence failures leave the session in COMMITTING with no receipt, so byte-identical retries recover.
- **Unknown-field policy decided**: v3 transport parsing stays tolerant of unknown JSON fields. Security does not depend on JSON strictness — every signed/MACed byte sequence comes from the canonical binary codec over a frozen field list, and version + ciphersuite are transcript-bound; extensions must arrive via a new negotiated protocol version. The design doc (§9.2) was amended accordingly and a test locks the tolerant behavior.
- New tests: real-HTTP unknown-field tolerance, reverse-direction downgrade guard (initiator role), v2 compatibility without v3 sessions, and a store-level commit/reject race test. Full `app:desktopTest` (1278 tests), `core:desktopTest`, `core:nativeAppTest`, and `ktlintCheck` pass.

PR #4683, opened 2026-07-29
:sparkles: add pairing v2 (SAS) and v3 (SPAKE2 PIN) support to the browser extension

### LINKED ISSUE #4682: Browser extension only supports legacy v1 bearer-token pairing
## Problem

The Chrome extension pairs with the desktop exclusively through the legacy v1 trust route (`POST /sync/trust` with a 6-digit bearer token). It neither implements the v2 SAS/ECDH handshake (`/sync/trust/v2/*`, shipped in desktop 2.0.0) nor the v3 SPAKE2+PIN protocol (`/sync/pairing/v3/*`), and its `AppInfo` does not even carry a `pairingVersion` field, so the desktop treats it as a pre-v2 peer.

This makes the extension the last active v1 client and blocks removing the v1 pairing protocol from the desktop server.

## Goal

- Implement v2 (SAS compare over ECDH key exchange) and v3 (SPAKE2 + PIN, initiator role) pairing in the extension.
- Negotiate the pairing mode from the desktop's advertised `pairingVersion` (1 = bearer token fallback, 2 = SAS, 3 = PIN), and advertise the extension's own pairing capability.
- Reuse the shared Kotlin `core` crypto (compiled to JS) instead of reimplementing any cryptography in TypeScript; the SPAKE2 P-256 backend for JS should follow the ADR D7 reviewed-library rule.
- Register the non-mDNS-discoverable extension on the v3 commit path the same way the v2 confirm route does (`crosspaste-sync-info` header).

## Notes

- Production desktops still advertise `pairingVersion = 2`, so v3 in the extension only activates against development builds until the desktop v3 rollout; the v2 path is the immediately user-visible improvement.
- Removing v1 server-side remains a follow-up, gated on this work plus mobile client verification.

guiyanakuang:
Closes #4682

## What

The extension previously only implemented v1 bearer-token pairing. This PR adds the v2 (SAS/ECDH) and v3 (SPAKE2 + PIN) protocols, negotiated from the desktop's advertised `pairingVersion`, and makes the extension advertise `pairingVersion = 3` itself.

### core (Kotlin Multiplatform)
- **`PairingV3InitiatorEngine`** (commonMain): transport-free v3 initiator whose canonical encodings, transcript construction, key schedule, and MAC contexts are byte-aligned with the desktop `PairingProtocolV3Service` initiator path; reusable by future mobile web surfaces.
- **`NoblePakeEcOps`** (jsMain): SPAKE2 P-256 backend over the audited `@noble/curves` library — per ADR D7's reviewed-library-per-platform rule (BouncyCastle on JVM, OpenSSL on native, noble on JS). No hand-rolled EC arithmetic.
- **`CrossPasteApi`** exports the v2 primitives (key exchange build/verify, `computeSAS`, trust confirm) and a `JsPairingV3Initiator` session facade for TypeScript.

### desktop
- `/sync/pairing/v3/commit` now reads the `crosspaste-sync-info` header (mirroring the v2 confirm route) so non-discoverable initiators self-register.
- `trustSyncInfo` drops a client-supplied SyncInfo whose `appInstanceId` does not match the authenticated caller (defense in depth, applies to both v2 and v3 registration paths).

### web extension
- Pairing-mode negotiation (`selectPairingMode`): 1 = bearer token (pre-2.0 desktops), 2 = SAS compare with local recomputation (mismatch aborts as possible MITM), 3 = SPAKE2 proof with the PIN shown on the desktop's device card.
- v3 recovery: a wrong PIN, an expired PIN, or any proof-send failure refreshes the offer (byte-identical intent resend) so the desktop rotates a fresh PIN and the attempt stays retryable.
- Mode-aware dialog copy and error messages in 10 locales; `@noble/curves` added to the extension dependency set (lockfile-managed, vite dedupe like `@js-joda/core`).

## Tests
- Initiator engine: 9 scripted-acceptor tests (real ECDSA both sides, acceptor re-derives every MAC/signature independently) running on **JVM, JS (Chrome headless), and native** targets.
- RFC 9382 Appendix B vectors refactored into a shared abstract suite now gating **both** the BouncyCastle and noble backends byte-for-byte; noble provider round-trip tests on JS.
- vitest: v2/v3 API tests with real WebCrypto (signature verification, SAS symmetry, tamper rejection) — 97/97.
- Full matrix green: `:core:build` (all targets), `:app:desktopTest`, `ktlintCheck`, `:web:build`, `npm test`.

## Review summary

An independent strict review (protocol alignment, noble-vs-BouncyCastle semantics, orchestration state machine, downgrade surface, test validity) found no P0/P1. Three P2 findings:
1. **Fixed**: `trustSyncInfo` identity check (above).
2. **Fixed**: v3 proof network-failure recovery (above).
3. **Accepted as a known design premise**: the pairing mode is selected from the unauthenticated `/sync/syncInfo` broadcast, so an active MITM could force the v1 path. This does not bypass verification — all three modes gate trust on an out-of-band secret the user reads off the desktop screen, and it matches how desktop peers negotiate today. When v1 is eventually removed, capability negotiation should move under transcript coverage.

## Rollout note

Production desktops still advertise `pairingVersion = 2`, so the extension uses v2 against current releases; v3 activates only against development builds with the v3 interop flag until the desktop v3 rollout. Real-device verification of both flows against a dev desktop is recommended before release.


guiyanakuang:
## Second review round — 5 P1 / 2 P2, all addressed in b86c71ac

| Finding | Fix |
|---|---|
| **P1** v2 double exchange left the desktop SAS overlay open forever (`startRefresh` counted twice, confirm released once) | Exactly one exchange per confirm: the connect-time exchange response is reused at pair time; a single re-exchange happens only on `EXCHANGE_TIMEOUT` (60s pending TTL), each one balanced by its confirm |
| **P1** re-pair could bind a different device's keys to the old record when the address changed hands | `handleRePair` now hard-fails with `device_identity_changed` when the answering `appInstanceId` differs from the stored one (v1/v2 servers don't verify `targetAppInstanceId`, so the client must) |
| **P1** concurrent CONNECTs could cross-bind identity and keys (global `connectingState` re-read after awaits) | `handlePair` snapshots the attempt; every step including `completePairing` operates on the snapshot and refuses to persist if the attempt was superseded |
| **P1** v3 network-failure recovery incomplete (proof-accepted-but-response-lost; no commit retry → one-sided trust on ACK loss) | Commit now retries 3× with deterministic bytes (idempotent, mirrors the desktop initiator). Unrecoverable proof/commit failures cancel the dead session and start a fresh one, so the next attempt converges both sides (same long-term keys get re-persisted) |
| **P1** extension v3 bypassed the production fail-closed gate (noble backend is not constant-time, ADR D7/Phase 0) | New `PAIRING_V3_ENABLED` gate: dev builds or `VITE_PAIRING_V3_INTEROP=true` only. Production builds cap negotiation at v2 and advertise `pairingVersion = 2`, mirroring the desktop's rollout gating |
| **P2** stale digits re-submitted after a wrong/expired PIN | `TokenInput` remounts empty and focuses the first cell after every failed attempt |
| **P2** cancel/failure paths never released the v3 session (desktop PIN card + downgrade guard lingered to TTL) | Closing the dialog sends `CANCEL_CONNECT` → best-effort `POST /sync/pairing/v3/cancel` + local key-material destroy; session replacement paths do the same |

Full verification re-run: `tsc`, `vite build`, vitest 98/98, `ktlintCheck` — green.

**Note for a possible follow-up (desktop-desktop, pre-existing):** the desktop's own v2 initiator also performs two exchanges per pairing (warm-up `exchangeKeysForPairing` + the real one inside `trustBySasCode`), which produces the same `refreshCounter` +2/−1 arithmetic on the responder. Worth checking whether the responder's token overlay relies on manual dismissal to drain the counter.

guiyanakuang:
## Third review round — 2 P1 / 3 P2, all addressed in 90bc4671

| Finding | Fix |
|---|---|
| **P1** v2 refresh counter still leaked (expired pending evicted without `stopRefresh`; re-exchange added another count; cancel paths couldn't release v1/v2) | Fixed at the root, desktop-side: `PendingKeyExchangeStore.put` now reports replacement and `lookup` distinguishes Live/Expired/None. The exchange route releases the superseded entry's count before starting a new one; the confirm route releases an expired entry's count. Every stored exchange now holds exactly one count and one confirm drains to zero → the SAS overlay auto-closes. **This also fixes the desktop-to-desktop warm-up double exchange (#4684's main path).** Remaining caveat: an abandoned exchange (dialog cancelled, never confirmed) still parks one count until the next exchange from that peer replaces it or the user closes the overlay — pre-v2-era behavior, there is no client-facing release endpoint; noted in #4684 |
| **P1** cancelling mid confirm/commit produced one-sided trust | `finalizing` flag on the attempt: CANCEL_CONNECT is refused while the desktop-persisting round-trip (v1 trust / v2 confirm / v3 commit) is in flight, and the dialog's close buttons are disabled while `loading` — the attempt always settles via completePairing or its failure path |
| **P2** v3 hard failures (proof response invalid / commit refused / ACK invalid) left an unusable engine behind a retryable-looking error | All three paths now best-effort replace the dead session (`restartV3Session`), so the next attempt starts clean with a fresh PIN |
| **P2** concurrent CONNECTs had no attempt generation (late failure clobbered newer state; same-target reconnects skipped release) | `beginPairing` snapshots the installed attempt at entry, aborts with "superseded" if a newer CONNECT won the preflight race (releasing its own v3 session), installs without intervening awaits, and unconditionally releases the replaced attempt — same target included. Late CONNECT failures no longer touch the global state |
| **P2** switching side-panel tabs skipped cancellation | `DevicesView` unmount cleanup fires CANCEL_CONNECT (refused while finalizing, no-op when idle) |

Verification: `:app:desktopTest` (incl. new `PendingKeyExchangeStoreTest` covering Live/Expired/None and replacement reporting), `ktlintCheck`, web `tsc` + `vite build` + vitest 98/98 — all green.

guiyanakuang:
## Fourth review round — 1 P1 / 3 P2, all addressed in 7c7a08e7

| Finding | Fix |
|---|---|
| **P1** cancelling a v2 attempt never released the desktop's pending exchange, and pending-exchange resources (refresh count, verifier, SAS view) had no unified owner across routes | Ownership unified desktop-side: `releasePendingKeyExchange` (remove entry + remove verifier + release refresh count) is the single release path, used by confirm success (now **unconditional** — the stored entry owns the count, not `showToken.value`), confirm expiry (now also removes the verifier), the v3 intent takeover, the v2-during-v3 rejection, and a **new `POST /sync/trust/v2/cancel` route** the extension calls from `releaseAttempt` (best-effort; pre-route desktops fail silently and fall back to the old manual-dismiss behavior). `PendingKeyExchangeStore.lookup` no longer lazily evicts, so ownership stays traceable until release |
| **P2** a cancel during finalizing was permanently lost | `cancelRequested` latch: the failure path honours it (cleanup, and `restartV3Session` skips creating a session the user no longer wants); successful completion ignores it |
| **P2** after a wrong PIN the refresh could race the server's async generation rotation and return the old offer, burning the freshly displayed PIN | `refreshOfferForNewGeneration`: the intent re-POST polls (5×200 ms) until `tokenGeneration` increases, falling back to the last offer on timeout |
| **P2** connect/cancel were global with no attempt id (multi-panel cross-cancel; early-started-late-arriving CONNECT could clobber a newer one) | CONNECT/REPAIR assign a monotonic attempt id at ENTRY and install aborts whenever a newer request has started, independent of network timing; PAIR and CANCEL_CONNECT carry the id (threaded through use-connection → DevicesView) so a stale dialog or another side panel can only act on its own attempt |

Remaining accepted gap: a v2 exchange abandoned against a desktop that predates the cancel route still parks one refresh count until the peer re-exchanges or the overlay is dismissed manually — inherent to old servers, converges as desktops update.

Verification: `:app:desktopTest` (store test updated for pure lookup + reporting remove), `ktlintCheck`, web `tsc` + `vite build` + vitest 98/98 — green.

PR #4769, opened 2026-08-14
:sparkles: serve a local /cli API over a unix domain socket and make the CLI a pure HTTP client

### LINKED ISSUE #4768: CLI: serve a local /cli API over a Unix domain socket and drop the CLI's direct database access
Part of the CLI completion effort (P1, follows #4766).

## Problem

The CLI currently opens the app's SQLite database directly and shells out to platform clipboard tools. This has fundamental problems:

- `status` always prints `Running` — it only reads the DB and never probes the app.
- `delete` / `config set` write the DB / `appConfig.json` behind a running app's back: in-process flows and caches never see the change, and `config set` is silently overwritten on the app's next config flush.
- `delete` runs with a no-op task submitter, so cleanup tasks for file-backed pastes never fire.
- `copy` pipes through `pbcopy`/`xclip`/`clip.exe`: xclip is not installed by default, there is no Wayland fallback, and `clip.exe` reads stdin in the OEM code page so non-ASCII text is garbled. The copied text also bypasses the app entirely and never enters history.
- The CLI links against system sqlite3, which makes cross-compiling (notably linuxArm64) fragile and drags `:shared` plus a hand-maintained 35-field mirror of `AppConfig` into the binary.

## Approach

Make every machine's peer process (today: the desktop app) the single owner of the database and clipboard, and turn the CLI into a thin HTTP client:

1. **App side**: an independent CIO `embeddedServer` bound to a Unix domain socket serves a full `/cli/*` route face (status, paste, history, search, devices, tags incl. create/delete, config get/set, copy, delete). The socket has no TCP port (unreachable from the network) and lives in a per-user 0700 directory with 0600 permissions, so other local users are locked out by the kernel; same-user processes are inside the trust boundary anyway.
2. **Discovery**: on readiness the app atomically writes `cli-endpoint.json` (pid, socketPath, apiVersion, appInstanceId) into its user data path and removes it on graceful shutdown. The CLI never guesses the socket path (macOS `sun_path` is limited to ~104 bytes, so the socket cannot live under a custom storage path).
3. **CLI side**: all commands go through the endpoint file + Unix socket; the database module, DAO wiring, config mirror, popen clipboard pipes, and the sqlite3 link are deleted. `copy` now writes the system clipboard **and** history via the app.

The CLI binary becomes pure Kotlin (Clikt + Ktor CIO client), which also removes the linuxArm64 sqlite cross-compilation risk ahead of packaging work.

guiyanakuang:
Closes #4768.

## Summary

- **App side** (`app/src/desktopMain/kotlin/com/crosspaste/cli/`):
  - `CliServer`: independent CIO `embeddedServer` on a Unix domain socket (`ktor-server-cio` added; the Netty sync server is untouched). Socket at `{java.io.tmpdir}/crosspaste-<user>/cli.sock`, kept short for the macOS `sun_path` ≤104-byte limit and length-checked; the directory is created 0700 with an ownership check (blocks another local user pre-creating it on shared `/tmp` and planting a fake socket), the socket file is chmod 0600 after bind, and stale socket files are removed before start. Non-production builds use suffixed socket names (`cli-development.sock`, …) so a dev instance cannot hijack the production socket.
  - `cliRouting`: full `/cli` route face — `GET status | paste/latest | paste/{id} | history | search | devices | tags | config`, `POST tags | copy`, `PUT config`, `DELETE paste/{id} | tags/{id}`. `copy` stores the paste (source `CLI`, real task pipeline) and writes the system clipboard; `delete` runs with the real task submitter so cleanup tasks fire and in-app flows refresh; `config` goes through `DesktopConfigManager` so memory and disk stay consistent, with key validation and type-checked value parsing.
  - `CliEndpointFile`: atomic (temp+rename, 0600) write of `cli-endpoint.json` (pid / socketPath / apiVersion / appInstanceId) on readiness, deleted on graceful shutdown — the single source of truth the CLI uses for discovery, resolved through `UserDataPathProvider` so a custom `storagePath` is honored.
- **CLI side** (`:cli`): every command now goes endpoint file → Unix socket HTTP. Deleted: `cliDatabaseModule`, DAO wiring, `NoOpServices`, `CliAdapters` (incl. the 35-field `AppConfig` mirror), `ClipboardPipe` (popen/`clip.exe`), Koin, the `-lsqlite3` link and the `:shared` dependency — the binary is now pure Kotlin (Clikt + Ktor CIO client), which also removes the linuxArm64 sqlite cross-compilation risk. `CliConfigReader` shrinks to a two-field `appConfig.json` projection used only to locate the endpoint file. An API-version mismatch prints a stderr warning without failing. `tags create/delete` are preserved via the new tag routes.
- **Contract**: DTOs are defined on both sides (K/Native and JVM cannot share a source set) with `ignoreUnknownKeys`; `CliDtoContractTest` pins the serialized field sets.

## Tests

- `CliRoutingTest` — all routes via `testApplication` + MockK (filters, validation, error mapping, copy/delete semantics).
- `CliServerTest` — real Unix-socket end-to-end: raw HTTP over `SocketChannel`, socket 0600, endpoint file lifecycle across start/stop.
- `CliEndpointFileTest` — atomic write, owner-only permissions, replace, delete.
- `CliDtoContractTest` — wire-contract pinning.

`./gradlew :app:desktopTest` (full suite) and `:cli:linkDebugExecutableCliNative` pass.

## Manual verification (macOS)

Against a real dev app instance (with a custom `storagePath`): `curl --unix-socket` works on all routes; the CLI binary completed a full loop — `status`, `copy` (verified via `pbpaste` and history entry with source `CLI`), `paste`, `search`, `tags`, `delete`. Error paths verified with a fake endpoint: missing endpoint file / dead socket → "CrossPaste is not running" (exit 1), API version mismatch → warning only.

## Notes

- Windows cross-user connection rejection relies on the per-user `%TEMP%` ACL; real-machine verification is queued for the packaging phase (P3) checklist.
- Auto-launch of the app when it is not running is the next phase (P2); the CLI currently reports "not running" with exit 1.


guiyanakuang:
Follow-up hardening after an independent review pass (0daed3fb):

- **Fail closed on foreign-owned socket directories**: the ownership check now runs unconditionally on POSIX filesystems instead of only when chmod succeeds. Previously, a directory pre-created by another local user on shared /tmp made chmod fail with EPERM, which silently skipped the ownership check — exactly the scenario the check exists for. A world-writable but self-owned directory is tightened to 0700 (covered by a new test).
- **Start/stop race**: CliServer start() is launched fire-and-forget at app startup; a mutex plus a stop-requested flag now keep a fast shutdown from racing it and leaving a stale socket/endpoint file. Failed starts also clean up the socket file.
- Blocking DAO calls in route handlers moved onto ioDispatcher; AppReadinessChecker reuses one HttpClient across polls.

Kept as-is intentionally: `/cli/history` reports `total` as the unfiltered active count even when type/tag filters are set — this matches the previous CLI behavior ("N of <all> pastes").

guiyanakuang:
Second review round (e3e5174a) — all four findings confirmed and fixed:

1. **`PUT /cli/config` false success**: `DesktopAppConfig.copy()` silently dropped `enableDiscovery` (a field-by-field scan showed it was the only key missing from the 53-field mapping — added), and `updateConfig` rolls back silently when persisting fails. The route now verifies the observed value after the update and answers 500 "was not applied" on mismatch.
2. **Storage keys**: `storagePath` / `useDefaultStoragePath` are rejected with 400 — changing them raw would bypass the app's migration flow and desynchronize endpoint-file discovery from the running server. A future migration API can lift this.
3. **`POST /cli/copy` lifecycle**: now creates a LOADING row and finishes it via `releaseLocalPasteData`, so process plugins, dedup (`markDeleteSameHash`) and the sync/rendering tasks all run — CLI copies now sync to paired devices like any local copy. A failed system-clipboard write returns 500 instead of a false "Copied".
4. **Hardening**: the CLI server's request decoder sets `ignoreUnknownKeys` (add-only DTO contract); the endpoint file is written through a randomly named temp file created with owner-only permissions at open time (no umask window, no fixed reusable temp name).

Re-verified against a live dev app: copy → history entry + `pbpaste` + delete; `config set enableDiscovery false/true` now actually applies; `config set storagePath` is rejected.

PR #4771, opened 2026-08-14
:sparkles: probe app liveness and auto-start the desktop app from the CLI

### LINKED ISSUE #4770: CLI: probe app liveness and auto-start the desktop app before running commands
Part of the CLI completion effort (P2, follows #4768).

## Problem

With the `/cli` Unix-domain-socket API in place (#4768), every CLI command depends on the local peer process (the desktop app) being up — but nothing checks or ensures that:

- If the app is not running, every command fails with a bare "CrossPaste is not running" error and generic exit code 1; the user has to start the app manually and re-run.
- If the app is still starting up (endpoint file written, socket not yet accepting), commands fail instead of waiting a moment.
- A stale `cli-endpoint.json` left behind by a crashed app (SIGKILL skips graceful cleanup) is indistinguishable from a starting app.
- `status` can only report "running or error"; there is no honest three-state view.
- The launcher/readiness helpers written earlier (`AppLauncher`, `AppAutoStarter`, `AppReadinessChecker`) were never wired to any command.

## Approach

**D5 liveness sequence** (`AppReadinessChecker.probe()`), run as a unified pre-flight for every command:

1. `cli-endpoint.json` missing/unreadable → **not running**.
2. `GET /cli/status` over the declared socket answers 200 → **running**.
3. Otherwise the recorded pid disambiguates: pid alive (POSIX `kill(pid, 0)`, Windows `OpenProcess` + `GetExitCodeProcess`) → **starting**, wait for readiness (≤30 s); pid dead → crash leftover, treat as **not running**.

**Auto-start with explicit consent** when the app is not running:

- Interactive TTY → `Start it now? [Y/n]` prompt (written to stderr; stdout stays pipe-clean).
- Non-interactive callers never block on input: fail fast with **exit code 3** ("app not running") and a hint to use `--start`.
- Root-level `--start` / `--no-start` force the answer either way (tri-state, default = ask).
- Headless Linux (no `DISPLAY`/`WAYLAND_DISPLAY`) gets a clear error instead of a doomed launch attempt (headless daemon support is P6).
- The launcher now pre-checks that the packaged start script exists, so a broken/dev install fails immediately instead of timing out after 30 s.

**Mid-command resilience**: if the app goes away while a command is running, the connection failure proves the request never reached it, so the CLI re-runs the liveness sequence (including the consent flow) and retries the request once. Requests that were delivered and then failed are not retried.

**`status` is special**: it never triggers auto-start, reports the actual three-state (`Running` / `Starting` / `Not running`), emits `running: bool` + `state` in `--json`, and exits 3 unless the app is ready to serve.

The pid check is the only platform-specific code; it lives in `posixNativeMain` / `mingwNativeMain` source dirs selected by the compilation target (not the host, so `-Pcli.target` cross-builds stay correct).

## Verification

All five K/N targets compile. Exercised on real machines (macOS binary + scripted UDS server + `expect` PTY; Docker Ubuntu for the headless branch):

- three probe states, live/dead pid disambiguation, stale-endpoint handling
- TTY prompt accept/decline/invalid-input reprompt; prompt on stderr with stdout redirected
- non-TTY fail-fast exit 3, `--no-start`, `--start` full launch chain via the packaged start script
- starting-state wait-then-proceed; mid-command server death → clean re-probe path
- API version mismatch warning; stdout purity under `--start`

guiyanakuang:
Closes #4770. Part of the CLI completion effort (P2, follows #4769).

## Summary

- **D5 liveness probe** (`AppReadinessChecker.probe()`): endpoint file → HTTP over the Unix socket → pid disambiguation (POSIX `kill(pid,0)` / Windows `OpenProcess`+`GetExitCodeProcess`), yielding an honest three-state result (running / starting / not running). The pid check is the only platform-specific code, in `posixNativeMain` / `mingwNativeMain` srcDirs keyed on the compilation target (not the host, so `-Pcli.target` cross-builds pick the right one).
- **Unified pre-flight in `runCli`** for every command: running → proceed; starting → wait for readiness (≤30 s); not running → consent flow, then launch via the previously unwired `AppLauncher`/`AppAutoStarter` and wait.
- **Consent flow**: interactive TTY asks `Start it now? [Y/n]` (prompt hand-rolled to stderr so stdout stays pipe-clean even with stdout redirected); non-interactive callers never block — exit code **3** ("app not running") with a `--start` hint. Root options `--start` / `--no-start` force the answer (tri-state, default = ask). Headless Linux (no `DISPLAY`/`WAYLAND_DISPLAY`) gets a clear error instead of a doomed launch (headless daemon is P6).
- **Mid-command resilience**: a connection failure means the request never reached the app, so the CLI re-probes (including consent) and retries once; requests that were delivered and then failed exit 1 without retry, so non-idempotent calls are never doubled.
- **`status` rework**: never auto-starts; prints `Running` / `Starting` / `Not running`; `--json` gains `running: bool` + `state`; exits 3 unless running.
- **Launcher hardening**: pre-check that the packaged start script exists — the backgrounded shell can't report a missing script, which previously meant a silent 30 s timeout.
- All progress/prompt messages go to stderr; stdout carries command output only.

## Verification

All five K/N targets compile (`macosArm64` native + `mingwX64`/`linuxX64`/`linuxArm64` cross). Exercised for real (macOS binary + scripted UDS server + `expect` PTY; Docker Ubuntu for headless):

- three probe states; live/dead pid disambiguation; stale endpoint (dead pid) treated as not running
- TTY prompt accept / decline / invalid-input reprompt; prompt lands on stderr with stdout redirected to a file (file stays empty)
- non-TTY fail-fast exit 3 with `--start` hint; `--no-start`; `--start` full launch chain via a packaged-layout start script ending in the original command succeeding
- starting state (pid alive, socket not yet up): waits ~3 s until the socket appears, then proceeds
- mid-command server death: re-probe path exits cleanly (3) instead of a raw error
- API version mismatch → single stderr warning, command still succeeds; Linux headless error branch in Docker

Real-app launch on a packaged install needs a build that contains the `/cli` server (first release including #4769), covered by the P3 packaging phase's release smoke test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,7 +113,7 @@ jobs:
       # Pure Kotlin since the CLI went all-HTTP; only needs the Kotlin/Native
       # toolchain already set up above.
       - name: Build CLI native binary
-        run: ./gradlew :cli:ktlintCheck :cli:linkDebugExecutableCliNative
+        run: ./gradlew :cli:ktlintCheck :cli:cliNativeTest :cli:linkDebugExecutableCliNative
 
   web-build:
     needs: changes
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -117,9 +117,18 @@ kotlin {
         }
     }
 
+    // Platform-split sources keyed on the selected target (not the host,
+    // which would break -Pcli.target cross-compilation)
+    val isMingwTarget = cliTarget == "mingwX64" || (cliTarget == null && isMingwX64)
+
     sourceSets {
         val cliNativeMain by getting {
             kotlin.srcDir(generateCliVersion)
+            if (isMingwTarget) {
+                kotlin.srcDir("src/mingwNativeMain/kotlin")
+            } else {
+                kotlin.srcDir("src/posixNativeMain/kotlin")
+            }
             dependencies {
                 implementation(libs.clikt)
                 implementation(libs.kotlinx.coroutines.core)
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -138,5 +138,14 @@ kotlin {
                 implementation(libs.okio)
             }
         }
+        val cliNativeTest by getting {
+            if (!isMingwTarget) {
+                kotlin.srcDir("src/posixNativeTest/kotlin")
+            }
+            dependencies {
+                implementation(libs.kotlin.test)
+                implementation(libs.kotlinx.coroutines.test)
+            }
+        }
     }
 }
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -114,6 +114,12 @@ kotlin {
                     linkerOpts("--allow-multiple-definition")
                 }
             }
+            // The test binary needs the same Clikt duplicate-symbol workaround;
+            // GNU ld only hits it when konan compiler caches are enabled (host
+            // builds), which is why cross-links from macOS don't reproduce it
+            if (!isMacosTarget) {
+                getTest("debug").linkerOpts("--allow-multiple-definition")
+            }
         }
     }
 
--- a/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/CrossPasteCommand.kt
+++ b/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/CrossPasteCommand.kt
@@ -17,6 +17,7 @@ import com.github.ajalt.clikt.core.obj
 import com.github.ajalt.clikt.core.subcommands
 import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.options.switch
 
 class CrossPasteCommand : CliktCommand(name = "crosspaste") {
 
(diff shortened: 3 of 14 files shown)


guiyanakuang:
Addressed the review findings in 4e06e68:

- **BLOCKER — unbounded wait**: `waitForAppReady` now runs under a hard `withTimeoutOrNull(30s)` deadline, and each probe uses the D5-specified 500 ms request timeout. A server that accepts connections but never answers can no longer stretch the wait to ~150 s; a virtual-time test pins the 30 s cap even with 2 s-per-probe stalls.
- **appInstanceId verification**: a 200 from the socket only counts as RUNNING when the response's `appInstanceId` matches the endpoint file (`matchesEndpointInstance`). A stale file pointing at a different instance now degrades to STARTING/NOT_RUNNING, and the wait loop re-reads the file every round so a restarted instance converges once it rewrites the file. Verified against a live socket: mismatched endpoint → `Starting`, matching → `Running`.
- **Version-warning regressions**: `runCli` tracks whether the warning was actually emitted (a retry client is still checked), and `status` re-gained the warning via the shared `warnOnApiVersionMismatch` helper.
- **`status` exit-code race**: a non-`AppNotRunningException` failure now re-probes; unless the app demonstrably serves (RUNNING), status prints the real state and exits 3. Exit 1 is reserved for a live app returning errors.
- **Test gap**: added `cliNativeTest` (20 tests, wired into the CI cli gate): three-state decision table incl. corrupted/stale endpoint files, the 30 s deadline (fast and slow probes, virtual time), mid-wait endpoint file re-read, instance-identity matching (mismatch, unknown fields, unparseable body), POSIX pid liveness (own pid, EPERM-as-alive via pid 1, invalid pids), and the exit-code-3 contract for non-interactive `paste`, `--no-start`, `status`, and `--json status` under an isolated HOME. The probe/pid seams in `AppReadinessChecker` are constructor-injectable for tests only.

PR #4776, opened 2026-08-15
:construction_worker: ship the CLI binary in all desktop packages

### LINKED ISSUE #4775: Ship the CLI binary in all desktop packages (CLI completion P3)
## Background

Phases P0–P2 of the CLI completion plan gave the desktop app a local `/cli` API over a unix domain socket (#4769) and made the CLI a pure HTTP thin client with liveness probing and auto-start (#4771). But no installer actually ships the CLI binary: it is not built by the release pipeline, not included in any package, and not reachable from a terminal after installing CrossPaste.

## Goal (P3: packaging and distribution)

Ship `crosspaste-cli` inside every desktop package and make it usable from a terminal right after installation:

- Build all five targets (`macosArm64`, `macosX64`, `linuxX64`, `linuxArm64`, `mingwX64`) in the release pipeline. Apple targets must be compiled on the macOS job; everything else cross-compiles on the Ubuntu job that runs Conveyor.
- Map the binaries into the packages via Conveyor:
  - macOS: `CrossPaste.app/Contents/bin/crosspaste-cli` (signed and notarized with the bundle);
  - Windows: `bin\crosspaste-cli.exe` next to `CrossPaste.exe`, plus an MSIX app execution alias `crosspaste-cli` so the command works in any terminal without PATH changes;
  - Linux: `/usr/lib/crosspaste/bin/crosspaste-cli` in the deb, with the `/usr/bin/crosspaste` symlink pointing at the CLI (the desktop entry launches the GUI via its absolute path and is unaffected).
- Inject the full release version from the git tag into the CLI build on both jobs (the version file is only rewritten later on the Ubuntu job).
- Keep the beta Linux workflow in sync.

## Notes

- On Windows, Conveyor hardcodes a `crosspaste.exe` execution alias that launches the GUI (via `updatecheck.exe`) and offers no config key to repoint it, so the CLI command on Windows is `crosspaste-cli`; on macOS/Linux the terminal command is `crosspaste`.
- The current single dynamic `-Pcli.target` Gradle setup can only build one target per invocation with a fixed task name, so it must be refactored into five static targets first.
- macOS PATH integration (offering a `/usr/local/bin/crosspaste` symlink from the app) is a separate follow-up.

guiyanakuang:
Closes #4775

## Summary

Ships the Kotlin/Native CLI inside every desktop package (P3 of the CLI completion plan, after #4767 / #4769 / #4771).

Core design (decision D6, revised during review): **the CLI is application payload, and where it lives inside a package is decoupled from how users reach it.** The binary is imported with plain per-arch Conveyor `inputs` like any other app file, so one mechanism covers every channel — deb, Linux tarball, AppImage, mac zip, Windows zip, and MSIX — with no repacking. The user-facing entry point is PATH, wired per channel at the most appropriate time.

### Gradle: five static targets

- `cli/build.gradle.kts` drops the single dynamic `-Pcli.target` target (one target per invocation, clashing output paths) and statically declares `macosArm64` / `macosX64` / `linuxX64` / `linuxArm64` / `mingwX64`. The existing source layout is reused as leaf source dirs per target, so no expect/actual split is needed.
- New `assembleDist<Target>` tasks stage release executables under `cli/dist/<target>/crosspaste-cli(.exe)`; `assembleDist` builds the host-buildable subset (local convenience, CI calls per-target tasks). `cliNativeTest` is a stable umbrella for the host target's test task.
- Linker-flag fix discovered by actually cross-linking: mingw links go through the clang++ driver, so the Clikt duplicate-symbol workaround must be passed as `-Wl,--allow-multiple-definition` there, while Linux targets keep the bare flag (konan invokes the linker directly for them).

### Packaging (`cli.conveyor.conf`)

New config included from `build.conveyor.conf` (release CI) and the beta workflow's generated conf — deliberately not from `conveyor.conf`, so a plain local `conveyor make app` keeps working without pre-building the CLI.

- **Placement** — per-arch `inputs`, landing in the JVM payload area of each platform: `lib/app/bin/crosspaste-cli` on Linux, `app\bin\crosspaste-cli.exe` on Windows, `Contents/Resources/bin/crosspaste-cli` on macOS. This reaches the Linux tarball (previously it could not carry the CLI at all: `root-inputs` are deb-only, and Conveyor has no package-root injection mechanism for tarballs) and, through it, the AppImage — `package_appimage.sh` only keeps a `test -x` regression check.
- **PATH entry points**:
  - Linux deb: the default launcher symlink is overridden so `/usr/bin/crosspaste` points at the CLI payload path (install time). The desktop entry launches the GUI via its absolute path and is unaffected. Tarball users wire PATH manually.
  - Windows MSIX: a second app execution alias `crosspaste-cli.exe` → `app\bin\crosspaste-cli.exe` is injected via `manifests.msix.extensions-xml`, so `crosspaste-cli` works in any terminal right after install. (Conveyor hardcodes the `crosspaste.exe` alias to the GUI launch path with no config key to repoint it, hence the separate name; the CLI's help/hints use a platform-aware command name so Windows users are never pointed at the GUI alias.)
  - macOS: a first-launch prompt offering a `/usr/local/bin/crosspaste` symlink is a follow-up PR; until then the binary is reachable inside the bundle.

### Direct app launch (start.sh decoupling)

`AppLauncher` no longer reuses `start.sh` / `start.bat` — their real job is the updater's wait-for-old-pid restart (`DesktopAppRestartService`), and they don't exist in every channel (the tarball has no `root-inputs`, hence no `start.sh`). The CLI now launches the app directly: `open <bundle>` on macOS, the Conveyor launcher `bin/crosspaste` on Linux, `start "" bin\CrossPaste.exe` on Windows. The launcher executable exists in every channel, so auto-start also works for tarball/zip installs. `CliAppPathProvider` derives paths from the payload-area location and exposes a single `resolveGuiLaunchTarget()`.

### Release workflow topology

- `build-macos` compiles both Apple targets and hands them to `build-ubuntu` as an artifact (same hand-off pattern as the Swift dylib; artifact instead of cache because the binaries change every release). Executable bits are restored after download — GitHub artifacts drop them.
- `build-ubuntu` cross-compiles `linuxX64` / `linuxArm64` / `mingwX64` before running Conveyor. Both jobs stamp the CLI with `-PcliVersion=${GITHUB_REF_NAME}` (the full tag), because `validateAndUpdateVersion.js` only rewrites the version file on the Ubuntu job after `build-macos` has already run.
- `build-beta-linux` builds the two Linux targets and includes the new Conveyor config.
- The PR CI gate now also release-links `linuxArm64` and `mingwX64`, guarding the cross-compilation path the release depends on.

## Verification

- All five targets compile from a macOS host; `:cli:cliNativeTest` passes (20 tests); ktlint clean; CI green including the Linux-host cross-links.
- Local Conveyor builds inspected for all channels: mac `.app` (CLI at `Contents/Resources/bin/`, auto-signed, bundle passes `codesign -v --strict --deep`, runnable), Linux tarball (CLI present with exec bit, launcher alongside), deb (payload binary + single overridden `/usr/bin/crosspaste` symlink), MSIX (`app\bin\crosspaste-cli.exe` + both execution aliases in `AppxManifest.xml`).
- Binary sizes: 5.8–7.7 MB per target; each package carries exactly one.
- Deferred to release smoke (documented in the release checklist): notarization over the bundled CLI, MSIX double-alias install validation on a real Windows machine, and the packaged-app auto-start chain from P2.

## Follow-ups (out of scope)

- macOS PATH integration (first-launch prompt to symlink into `/usr/local/bin`) — remaining P3 item, separate PR.
- P4 terminal experience (stdin copy, `--raw`, exit-code docs, CLI docs page).

### DIFF
--- a/.github/scripts/package_appimage.sh
+++ b/.github/scripts/package_appimage.sh
@@ -40,6 +40,20 @@ build_appimage() {
     # (e.g. crosspaste-1.2.4/) and extract contents directly into BUILD_DIR
     tar -xzf "$TAR_FILE" -C "$BUILD_DIR" --strip-components=1
 
+    # The Conveyor tarball never carries root-inputs (they are deb-only), so
+    # the CLI binary must be copied in from the staged dist explicitly or the
+    # AppImage would silently ship without it. The CI workflows build the
+    # Linux CLI targets before this script runs; set -e turns a missing
+    # binary into a hard failure instead of a silent drop.
+    local CLI_TARGET
+    case "$ARCH_TOKEN" in
+        amd64) CLI_TARGET="linuxX64" ;;
+        aarch64) CLI_TARGET="linuxArm64" ;;
+        *) echo "Error: no CLI target mapping for arch ${ARCH_TOKEN}"; exit 1 ;;
+    esac
+    install -m 755 "cli/dist/${CLI_TARGET}/crosspaste-cli" "$BUILD_DIR/bin/crosspaste-cli"
+    test -x "$BUILD_DIR/bin/crosspaste-cli"
+
     # Configure AppImage metadata (AppRun, Desktop file, Icon) inside a subshell
     # so the working directory stays at the workspace root for appimagetool.
     (
--- a/.github/scripts/package_appimage.sh
+++ b/.github/scripts/package_appimage.sh
@@ -40,19 +40,11 @@ build_appimage() {
     # (e.g. crosspaste-1.2.4/) and extract contents directly into BUILD_DIR
     tar -xzf "$TAR_FILE" -C "$BUILD_DIR" --strip-components=1
 
-    # The Conveyor tarball never carries root-inputs (they are deb-only), so
-    # the CLI binary must be copied in from the staged dist explicitly or the
-    # AppImage would silently ship without it. The CI workflows build the
-    # Linux CLI targets before this script runs; set -e turns a missing
-    # binary into a hard failure instead of a silent drop.
-    local CLI_TARGET
-    case "$ARCH_TOKEN" in
-        amd64) CLI_TARGET="linuxX64" ;;
-        aarch64) CLI_TARGET="linuxArm64" ;;
-        *) echo "Error: no CLI target mapping for arch ${ARCH_TOKEN}"; exit 1 ;;
-    esac
-    install -m 755 "cli/dist/${CLI_TARGET}/crosspaste-cli" "$BUILD_DIR/bin/crosspaste-cli"
-    test -x "$BUILD_DIR/bin/crosspaste-cli"
+    # The CLI arrives with the tarball itself (it is an app input landing in
+    # lib/app/bin/ — decision D6); just verify it is really there so a
+    # packaging regression fails the build instead of silently shipping an
+    # AppImage without the CLI.
+    test -x "$BUILD_DIR/lib/app/bin/crosspaste-cli"
 
     # Configure AppImage metadata (AppRun, Desktop file, Icon) inside a subshell
     # so the working directory stays at the workspace root for appimagetool.
--- a/.github/workflows/build-beta-linux.yml
+++ b/.github/workflows/build-beta-linux.yml
@@ -98,6 +98,17 @@ jobs:
       - name: Build with Gradle
         run: ./gradlew app:build
 
+      - name: Cache Kotlin/Native toolchain
+        uses: actions/cache@v5
+        with:
+          path: ~/.konan
+          key: konan-${{ runner.os }}-${{ hashFiles('gradle/libs.versions.toml') }}
+
+      - name: Build CLI binaries (Linux targets)
+        run: |
+          VERSION=$(grep '^version=' app/src/desktopMain/resources/crosspaste-version.properties | cut -d= -f2)
+          ./gradlew :cli:assembleDistLinuxX64 :cli:assembleDistLinuxArm64 "-PcliVersion=${VERSION}.${REVISION}"
+
       - name: Prepare Conveyor config (Linux only)
         run: |
           ./gradlew -q writeConveyorConfig -PappEnv=PRODUCTION
@@ -109,6 +120,7 @@ jobs:
           # so mac/windows signing secrets are never needed.
           cat > beta.conveyor.conf <<CONF
           include required("conveyor.conf")
+          include required("cli.conveyor.conf")
           app.machines = [ "linux.amd64.glibc", "linux.aarch64.glibc" ]
           app.revision = ${REVISION}
           CONF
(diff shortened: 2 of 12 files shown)


guiyanakuang:
Addressed both review findings:

- **AppImage missing the CLI (blocker)**: `package_appimage.sh` now installs `cli/dist/<linuxX64|linuxArm64>/crosspaste-cli` (mapped from the tarball arch token) into `AppDir/bin/` right after extraction, with an executable check; `set -e` turns a missing binary into a hard failure instead of a silent drop. The misleading "and AppImage built from it" comment in `cli.conveyor.conf` is corrected and now points at the script as the counterpart to keep in sync.
- **Windows command name (warning)**: the Clikt command name and the `--start` hint are now derived from a platform-aware `cliCommandName` (`crosspaste-cli` on Windows, `crosspaste` elsewhere), so usage/help output and the non-TTY hint match what is actually invocable on each platform.

`:cli:cliNativeTest` (20 tests) and ktlint pass locally; no test asserted the old literals.

guiyanakuang:
Design revision after review discussion (commit a5e4451): **the CLI is now imported as plain application payload, and the app is launched directly.**

Rationale: the package-root `bin/` placement needed a different Conveyor mechanism per OS (`bundle-extras` / `package-extras` / `root-inputs`), and Linux has no package-root mechanism that reaches the tarball — which forced the unpack-repack AppImage workaround. Where the binary lives inside the package is invisible to users; the user-facing entry point is PATH. Decoupling the two removes the whole problem:

- **Placement**: one mechanism everywhere — per-arch `inputs`, landing in the JVM payload area (`lib/app/bin/` on Linux, `app\bin\` on Windows, `Contents/Resources/bin/` on macOS). This reaches every channel (deb, **tarball — previously missing the CLI entirely**, AppImage via the tarball, mac zip, win zip, MSIX) with zero repacking; the AppImage injection from the previous commit is reverted to a plain `test -x` regression check.
- **PATH entry points** (unchanged in spirit, retargeted): deb symlink `/usr/bin/crosspaste` → payload path (install time); MSIX alias `crosspaste-cli` → `app\bin\crosspaste-cli.exe` (install time); macOS first-launch symlink prompt stays a follow-up PR.
- **Direct launch**: `AppLauncher` no longer reuses `start.sh`/`start.bat` (whose real job is the updater's wait-for-old-pid restart, and which never existed in the tarball). It now launches the app directly: `open <bundle>` on macOS, the Conveyor launcher `bin/crosspaste` on Linux, `start "" bin\CrossPaste.exe` on Windows. The launcher exists in every channel, so auto-start now also works for tarball/zip installs, and the earlier `app\bin\start.bat` path fix became unnecessary and is gone.

Local verification (Conveyor builds inspected for all three OSes): mac CLI at `Contents/Resources/bin/`, auto-signed, bundle verifies, runnable; Linux tarball now contains the CLI (exec bit preserved) alongside `bin/crosspaste`; deb symlink resolves to the payload path exactly once; MSIX carries `app\bin\crosspaste-cli.exe` with the alias Executable updated. `:cli:cliNativeTest` (20) and ktlint pass.

PR #4778, opened 2026-08-15
:sparkles: offer to install the crosspaste terminal command on macOS

### LINKED ISSUE #4777: Offer to install the crosspaste terminal command on macOS (CLI P3 follow-up)
## Background

Since #4776 every desktop package ships the CLI binary as application payload (design decision D6), and PATH integration is wired per channel at install time: the Linux deb creates the `/usr/bin/crosspaste` symlink and the Windows MSIX declares a `crosspaste-cli` execution alias. macOS is the remaining gap: an installer cannot touch PATH, so after installing CrossPaste.app the bundled CLI (`Contents/Resources/bin/crosspaste-cli`) is not reachable from a terminal.

## Goal

The VS Code model, as planned in D6:

- A one-time prompt after launch on macOS offering to install the `crosspaste` command — a symlink at `/usr/local/bin/crosspaste` pointing at the bundled CLI. Installing first tries a plain symlink (works where `/usr/local/bin` is user-writable) and falls back to an `osascript … with administrator privileges` credentials dialog, since the directory is root-owned or absent on stock installs.
- A settings entry to install, or later repair, the command (e.g. after the app moved and the symlink dangles), so declining the prompt is never final.
- Everything hidden on other platforms and in dev runs (no bundled CLI binary there).

## Notes

- The prompt must not stack with the accessibility permission dialog on first launch.
- Declining or completing the prompt persists a config flag so it never re-appears; the settings entry remains available.

guiyanakuang:
Closes #4777

## Summary

Completes the last piece of P3 (CLI packaging, design decision D6, after #4776): macOS PATH integration for the bundled CLI, following the VS Code model.

### `CliSymlinkService` (`app/src/desktopMain/.../cli/`)

- Maintains the `/usr/local/bin/crosspaste` symlink pointing at the bundled CLI. The binary is resolved as `pasteAppJarPath/bin/crosspaste-cli` — on a packaged macOS app `pasteAppJarPath` is `Contents/Resources`, verified against the shipped Conveyor bundle.
- State machine exposed as a `StateFlow`: `NOT_SUPPORTED` (not macOS, not a production install, or no bundled binary — i.e. dev runs), `INSTALLED` (symlink resolves to this bundle's CLI), `NOT_INSTALLED`, `NEEDS_REPAIR` (dangling link, link to another location, or a foreign file at the link path).
- `install()` first tries a plain `Files.createSymbolicLink` (works where `/usr/local/bin` is user-writable, e.g. Homebrew-on-Intel machines) and falls back to `osascript … with administrator privileges`, which shows the standard macOS credentials dialog. Paths travel as osascript argv and are quoted with AppleScript's `quoted form of`, so spaces/quotes in the install location can't break the shell command (exercised end-to-end with spaces in paths). A declined credentials dialog is recognized (AppleScript error `-128`) and reported as `CANCELLED` rather than an error.

### UI

- **One-time prompt** (`InstallCliDialog`): shown after launch when the command is not installed, gated by a new persisted flag `showInstallCliPrompt` which is cleared once the prompt is answered (installed or "Later") — the prompt never nags twice. It is suppressed while the accessibility permission dialog is up, so first-launch prompts don't stack. Cancelling the credentials dialog keeps the prompt open for another try.
- **Settings entry** (`CliToolSettingItem` in a macOS-only "Command Line" section): shows the current state and offers Install / Repair; this stays available forever, so declining the prompt is never final. Hidden entirely when the service reports `NOT_SUPPORTED`.
- i18n: 10 new keys across all 10 locales, alphabetically sorted.

## Verification

- `CliSymlinkServiceTest` (11 tests): every state including dangling links and foreign files at the link path, direct install and repair, escalation fallback when the link directory is not writable (injected escalation), and unsupported-mode failure.
- `DesktopAppConfigTest`: default value, `copy()` coverage, and legacy-JSON default for the new flag.
- Full `:app:desktopTest` suite and ktlint pass; the osascript quoting logic was additionally executed for real (sans admin clause) against paths containing spaces.
- Real prompt + credentials-dialog flow on a packaged app is covered by the release checklist (Phase 3.5 CLI smoke), same as the rest of the CLI packaging chain.


guiyanakuang:
Addressed review feedback on placement and refresh timing:

- **Refresh timing**: the symlink state is now probed once at app startup (next to `CliServer.start()` on the IO dispatcher), instead of lazily when a screen happens to compose — previously the extension entry's visibility depended on opening the settings screen first. The only other probe is the implicit re-check after `install()`.
- **Entry placement**: the install/repair row moved from the settings screen to the extension page (`ExtensionContentView`, macOS-only "Command Line" section under the MCP/OCR/source-control card); `SettingsContentView` is back to untouched. `CliToolSettingView` moved to the `ui.extension` package accordingly.
- The per-composition `refresh()` calls in the dialog host and the row were removed along with it.

Full `:app:desktopTest` + ktlint pass.

guiyanakuang:
All four review findings addressed (commit d072afa):

- **P1 App Translocation**: new `TRANSLOCATED` state, detected via the `/AppTranslocation/` path segment of the resolved CLI binary. Install is refused in this state, the first-launch prompt never fires for it (it only fires for `NOT_INSTALLED`), and the extension row swaps its subtitle to "move CrossPaste to the Applications folder and relaunch" with no action button. Covered by state + install-refusal tests.
- **P1 destructive repair**: new `CONFLICT` state for anything at the link path that is not a symlink (regular file or directory). `install()` now hard-guards: only `NOT_INSTALLED` and `NEEDS_REPAIR` are actionable, so neither the direct `deleteIfExists` nor the escalated shell command can ever run against a foreign file. The row explains the conflict via its subtitle and offers no button. Tests assert the foreign file survives an install attempt byte-for-byte.
- **P2 false success**: the escalated command changed from `ln -sf` to `mkdir -p … && rm -f <link> && ln -s <target> <link>` — `rm -f` removes a symlink without following it and refuses directories, so the link-into-directory trap is gone. Independently, `install()` now re-probes after the operation and only returns `SUCCESS` when the final state is verifiably `INSTALLED`; a test covers escalation "succeeding" without producing the link.
- **P2 prompt lifecycle**: the dialog's `FAILURE` branch no longer calls `doneAction()` — it shows the error notification, resets `installing`, and keeps the dialog (and the one-time flag) so the user can retry, matching the documented "installed or Later" semantics.

`CliSymlinkServiceTest` grew to 17 tests; full `:app:desktopTest` + ktlint pass.

guiyanakuang:
Added the command-line sub-page requested in review (commit 926ae9b):

- The extension page now shows a **Command Line navigation row** (in the same card as MCP / OCR / source control, macOS-only) instead of an inline install row; it opens a dedicated `CliTool` route under the extension graph.
- The sub-page shows, top to bottom: the **install/repair/status row** (previous `CliToolSettingItem`, unchanged behavior including the translocation/conflict states), the **bundled executable path** (selectable, monospace), and **Terminal Availability** — one row per login shell (zsh, bash) reporting whether `command -v crosspaste` resolves, with the resolved path shown when it does. The probes run through each shell as a login shell so the user's own PATH configuration is honored, are re-run whenever the symlink state changes (e.g. right after installing), and are guarded by a 5s timeout with the process killed on expiry.
- Route/title/back-target registrations follow the MCP sub-page pattern; 4 new i18n keys across all 10 locales.

Full `:app:desktopTest` + ktlint pass.

guiyanakuang:
Both findings addressed (commit 8345a4f):

- **P1 TOCTOU**: the symlink-only-delete invariant is now enforced at execution time in both paths, not just in the pre-check.
  - Direct path: extracted to `attemptDirectInstall()`, which re-checks the link path right before acting — a foreign file that appeared after the pre-check returns `FAILURE` **without escalating** (asking for admin credentials would not make destroying it acceptable), and only a verified symlink is ever deleted.
  - Escalated path: the shell command is now the `INSTALL_SHELL_TEMPLATE` constant — `mkdir -p …; if [ -L link ]; then rm -f link; fi; if [ -e link ] || [ -L link ]; then exit 40; fi; ln -s target link` — so the root-privileged `rm` can only ever remove a symlink, any other occupant bails out with exit 40 before touching anything, and plain `ln -s` keeps the no-follow guarantee. The template is exercised for real through `/bin/sh` in tests: foreign file bails out intact (exit 40), free path installs, wrong symlink is replaced, and a symlink-to-directory is replaced without creating anything inside the directory.
  - New regression tests also cover the direct path's execution-time refusal (calling `attemptDirectInstall()` with a foreign file in place, modeling the race).
- **P2 stale CONFLICT**: `CliToolContentView` now calls `refresh()` on every visit, so resolving a conflict (or removing the link) outside the app is noticed by re-entering the page — no restart needed. The recovery guidance also switched from the ellipsized subtitle parameter to a raw multi-line `subtitleContent`, so the full instruction is readable in every locale.

`CliSymlinkServiceTest` is now 22 tests; full `:app:desktopTest` + ktlint pass.

guiyanakuang:
Fixed (commit a1c7297):

- The probe command now tags its answer — `p=$(command -v crosspaste) && printf '__CROSSPASTE_RESOLVED__%s\n' "$p"` — and parsing extracts only the marker-prefixed line (`parseResolvedCommand`), so profile banners can never be mistaken for the executable path. The marker is only printed on success, which also removes the exit-code ambiguity.
- stdout is now drained concurrently with the wait: a profile printing more than the pipe capacity previously blocked the shell until the 5s timeout and read as "unavailable". Killing the process on timeout closes the stream, so the drain itself cannot hang.
- Tests: banner-noise parsing (including a path-looking line with no marker), a 200KB-banner probe through a real `/bin/sh` (fails without concurrent draining), and a command-not-found probe. `CliSymlinkServiceTest` is now 25 tests; full `:app:desktopTest` + ktlint pass.

guiyanakuang:
Addressed with a redesign rather than patches (commit a535ffa):

**The probe no longer performs any blocking read.** `resolveCommandIn` polls `available()` and scans line-by-line with a bounded buffer (64KB line cap — longer lines can't be the marker line, and a firehose profile can't balloon memory since scanned bytes are discarded):

- **Marker line → immediate return**, without waiting for EOF — a background job inheriting stdout (`sleep 30 &`) can hold the pipe open forever and is simply never waited on (it's also not ours to kill; we just stop reading).
- **Shell exit with nothing left to scan → immediate null**, again no EOF wait.
- **Timeout and cancellation ride on cancellable `delay`/`withTimeoutOrNull`** — no `runCatching` wraps the suspend path anymore, so `CancellationException` propagates untouched; only `ProcessBuilder.start()`'s `IOException` is caught specifically.
- **Cleanup in `finally` on every exit path** (success, timeout, cancellation): close our pipe end, `destroyForcibly()` a still-running shell.

New tests, both through a real `/bin/sh`: `sleep 30 & printf marker` returns the marker immediately (asserted under the 5s budget), and a hung foreground `sleep 30` returns null at an injected 1s timeout. The probe tests moved from `runTest` to `runBlocking` — the poll loop needs real time, and the virtual clock would race to the timeout before the shell produces output. `CliSymlinkServiceTest` is now 27 tests; full `:app:desktopTest` + ktlint pass.

PR #4780, opened 2026-08-15
:sparkles: improve the CLI terminal experience with pipes, formats, colors, and exit codes

### LINKED ISSUE #4779: CLI terminal experience: piped stdin, raw output, output formats, colors, and a documented exit-code contract
Phase 4 (terminal experience) of the CLI completion work. The CLI is functional as of the /cli UDS API (#4766), auto-start (#4770), and packaging (#4775), but it is not yet a good citizen in shell pipelines:

- \`copy\` only accepts an argument; \`git log | crosspaste copy\` is impossible.
- \`paste\` always prints decorated output; there is no way to pipe the bare content onward (\`crosspaste paste | pbcopy\`).
- \`history\`/\`search\` offer only the human table or full \`--json\`; bulk scripting (e.g. piping IDs to \`xargs\`) requires jq gymnastics.
- No colors, and no guarantee about TTY/\`NO_COLOR\` behavior.
- Exit codes are not fully specified: Clikt's default maps usage errors to 1, indistinguishable from generic failures, and the contract is not documented in \`--help\`.
- There is no user-facing CLI documentation.

Planned changes:

1. \`copy\` reads stdin when invoked with no argument and stdin is not an interactive terminal (never blocks on a TTY; bounded buffering).
2. \`paste --raw\` (\`-r\`) prints only the content, byte-faithfully, with \`--no-newline\` to suppress the trailing newline.
3. \`history\`/\`search\` gain \`--format table|json|id\` (\`id\` = one paste ID per line, for \`xargs\`).
4. Colors in human-readable output via Mordant, automatically disabled on non-TTY output and honoring \`NO_COLOR\`.
5. Exit-code contract documented in \`--help\` and enforced: 0 success, 1 error, 2 usage error, 3 app not running.
6. Documentation: a CLI page (installation/PATH per platform, command reference, piping examples, completion setup) plus a README mention.

guiyanakuang:
Closes #4779

Phase 4 (terminal experience) of the CLI completion work.

### `copy` reads piped stdin
- The `text` argument is now optional. With no argument, the command reads stdin to EOF when stdin is **not** an interactive terminal (`git log | crosspaste copy`); on an interactive TTY it fails with a usage error instead of blocking forever.
- Stdin is read via a `platform.posix.fread` loop (works on all five targets, including mingw), decoded as UTF-8 only after all chunks are concatenated (multibyte sequences split across read boundaries stay intact), and capped at 64 MiB so a runaway pipe cannot OOM the CLI (the app rejects oversized text pastes anyway; its default limit is 8 MB).
- Empty piped input is a usage error. The stdin reader is constructor-injected so tests never touch the process's real stdin.

### `paste --raw` / `--no-newline`
- `--raw` (`-r`) prints only the stored content, byte for byte, deliberately bypassing the Mordant terminal (which re-renders strings and would strip ANSI codes from the content on non-TTY output). `--no-newline` suppresses the trailing newline; using it without `--raw` is a usage error.
- A paste with no text content (file/image) reports an error on stderr and exits 1.

### `history` / `search --format table|json|id`
- An explicit `--format` wins; the global `--json` flag implies `json`; the default stays `table`. `id` prints one paste ID per line for piping (`crosspaste history --format id | xargs -n1 crosspaste delete`).

### Colors and TTY adaptation
- Human-readable output now uses a few Mordant colors: status state (green/yellow/red), paste types (cyan), timestamps/sizes (gray), tagged marker (yellow), device connection states. Mordant downsamples embedded ANSI to the terminal's capability, so colors are automatically stripped on non-TTY output and `NO_COLOR` is honored — verified against the real binary both ways (pseudo-TTY shows ANSI, redirect/`NO_COLOR` output is clean).
- Styling is applied after `padStart`/`padEnd` so ANSI escapes never break column alignment.

### Exit-code contract
- Documented in the root `--help` epilog: 0 success, 1 error, 2 usage error, 3 CrossPaste not running.
- Clikt 5's `UsageError` defaults to status code 1 (indistinguishable from generic errors), so `main` now hand-rolls parse + catch and remaps usage errors to 2, leaving every other `CliktError` (including `ProgramResult(3)`) untouched. The Clikt test harness bypasses `main`, so the remap is pinned by real-binary verification (unknown option, bad arguments, and `--no-newline` without `--raw` all exit 2).

### Documentation
- New `doc/en/CLI.md` and `doc/zh/CLI.md`: per-platform installation/PATH table, command reference, piping and scripting examples, exit codes, shell completion setup, color behavior.
- README (en/zh-CN) feature bullets linking to the CLI guide.

## Tests

- `CopyStdinTest` (5): interactive TTY refuses to read stdin, piped stdin is consumed and proceeds to the liveness check, empty stdin is a usage error, an explicit argument never reads stdin, oversized stdin fails with a clear error.
- `CliExitCodeTest` (+2): the help epilog documents the contract; `--no-newline` without `--raw` is rejected.
- `ListFormatTest` (3): format resolution precedence.
- Full CLI suite: 30 tests green; ktlint clean; all five targets compile (`macosArm64`, `macosX64`, `linuxX64`, `linuxArm64`, `mingwX64`).

## Manual verification (real binary, macOS)

- `printf 'hello' | crosspaste copy` reads stdin, then correctly enters the liveness flow (exit 3 with the `--start` hint when the app is stopped); `crosspaste copy < /dev/null` exits 2.
- `--bogus`, bad arguments, and `paste --no-newline` all exit 2; `status` with the app stopped exits 3.
- Colors appear under a pseudo-TTY and are absent on redirected output and with `NO_COLOR=1`.


guiyanakuang:
Pushed 6d8158f0: review found that `/cli/copy` returned a fake success for text over the app's configured `maxNonFilePasteSize` (default 8 MB) — `DiscardOversizedNonFilePlugin` empties the item list during release and the row gets mark-deleted, yet the handler still wrote the system clipboard and returned the (now deleted) id, so the content never entered history or synced.

`handleCopy` now validates `pasteItem.size` against the current `maxNonFilePasteSize` **before** creating the LOADING row and responds `413 Payload Too Large` with an explicit message (the CLI surfaces it on stderr and exits 1). The CLI-side 64 MiB stdin cap is unchanged — it only guards against unbounded buffering and is a separate layer from the app's configured limit. Added a `CliRoutingTest` case pinning the 413 and that no row is created and no clipboard write happens.

guiyanakuang:
Pushed 2c35253c: review found that `paste --raw` actually printed the *summary* — the detail endpoint filled `content` via `pasteItemReader.getSummary()`, so HTML/RTF came back converted to plain text, and file/image pastes returned a filename list instead of `null` (so `--raw` printed filenames and exited 0 rather than erroring as designed).

Per the agreed design:

- **`PasteItemReader.getRaw`** (core): new API alongside `getText` — HTML/RTF return their stored source markup; every other type falls back to `getSummary`. Implemented as an interface default method since reading the stored fields needs no platform parsing, which keeps the mobile copy of `core` at zero follow-up cost.
- **`/cli/paste` detail** gains an add-only `rawContent` field (`content` keeps its summary semantics); the DTO contract test and an HTML round-trip test (`content` = parsed text, `rawContent` = source markup) pin both.
- **CLI**: `--raw` now prints `rawContent`; the previous plain-text behavior moved to `--summary` (`-s`). The two are mutually exclusive and `--no-newline` requires one of them. File/image pastes print their filename list in both modes (their raw is their summary); the exit-1 error remains only for genuinely missing content.

Docs updated accordingly. Full suite green: 17 `CliRoutingTest` + 6 contract tests on the app side, 31 CLI-side tests; all five CLI targets and the core JS distribution compile; mutual-exclusion and `--no-newline` guards verified against the real binary (both exit 2).

guiyanakuang:
Pushed fdc13543: review found two silent-corruption risks in the stdin reader.

1. **Read errors were treated as EOF.** `readAllStdin` broke out of the loop on any `read <= 0` without checking `ferror(stdin)`, so a closed stdin was misreported as "stdin was empty" (exit 2), and a mid-stream read error would have silently copied truncated content. The loop is now followed by a `ferror` check that throws `StdinReadException` (message includes `strerror(errno)`), which `copy` reports on stderr and exits 1 — an error can never turn into a partial copy.
2. **Invalid UTF-8 was silently rewritten.** The default `decodeToString()` replaces bad sequences with U+FFFD; for a pipe tool, silently altering the input is worse than failing. Decoding is now strict (`throwOnInvalidSequence = true`) and invalid input fails with a clear message and exit 1. The decode step is split into an internal `decodeStdinUtf8` so the contract is unit-testable.

Both exceptions share a sealed `StdinException` base, so `copy` handles them at a single catch site. New tests: `StdinDecodeTest` (valid multibyte round-trip, invalid sequence, truncated multibyte tail) and a `CopyStdinTest` case for the read-error exit code — CLI suite now 35 green across all five targets. Verified against the real binary: `copy 0<&-` reports "failed to read stdin: Bad file descriptor" (exit 1), `printf '\xff\xfe' | copy` reports the UTF-8 error (exit 1), and a valid pipe still proceeds to the liveness check.

guiyanakuang:
Pushed 798e8e1a: two more review findings fixed.

1. **`--json` now works for every command, as the docs promise.** `copy`, `delete`, `config set`, `tags create`, and `tags delete` only ever printed plain text. Rather than weakening the documentation, the implementation was completed: `copy` emits the `CopyResponse` JSON (`{"id": …}`), `tags create` emits the created `TagSummary`, and the message-only commands (`delete`, `config set`, `tags delete`) share an `echoMessage` helper that passes the server's `{"message": …}` through as JSON under `--json`.
2. **Runtime usage errors now show the subcommand's usage, not the root help.** A `UsageError` thrown from `run()` carries no parse context, so Clikt rendered the full root command help for `paste --no-newline` and copy's stdin errors. A `usageError(message)` helper now attaches the throwing command's own context; verified against the real binary — `paste --no-newline` prints `Usage: crosspaste paste [<options>] [<id>]` followed by the error, still exit 2. (`CopyCommand` reads the CLI context via nullable `findObject` since tests exercise it standalone.)

Tests updated to pin the subcommand usage line in stderr (3 assertions); CLI suite 35 green, all five targets compile.

guiyanakuang:
Pushed e7be8ed3: review flagged a blocker in the previous raw-content change — the detail endpoint filled both `content` (summary) and `rawContent` unconditionally, and since `getRaw` falls back to `getSummary` for non-HTML/RTF types, a plain-text paste shipped its full content **twice** in every detail response. With the text limit configurable up to 64 MiB and the CLI buffering the whole body via `bodyAsText()` before deserializing, that doubled both the transfer and the client's peak memory even when `--raw` was never requested.

Raw is now opt-in:

- The detail routes accept `?includeRaw=true`; without it `rawContent` is `null` and only the summary is shipped (the field stays in the DTO, so the wire contract is unchanged / still add-only).
- The CLI adds the query parameter only when `--raw` is passed; `--summary`, the human view, and `--json` all use the default request.
- File/image pastes keep returning their filename list via `content`, per the agreed semantics.

`CliRoutingTest` pins both directions (default response must carry `rawContent = null`; `includeRaw=true` returns the HTML source). Full suite green on both sides; all five CLI targets compile.

guiyanakuang:
Pushed cead1062: review noted that the new `--raw` capability changed the wire contract while both sides still declared `CLI_API_VERSION = 1` — a new CLI talking to an older running app would get no version-mismatch warning, and `--raw` would misreport the paste as having no content (the old app simply ignores `?includeRaw=true`).

- `CLI_API_VERSION` bumped to **2** on both the app and CLI sides, with the version history noted at the constant (`2: paste detail supports ?includeRaw=true / rawContent`). The existing mismatch warning now fires for new-CLI/old-app pairs.
- `paste --raw` distinguishes the old-app case explicitly: when `rawContent` is missing but the summary is present, it reports "the running CrossPaste app does not support --raw yet; update (or restart) the app, or use --summary instead" and exits 1, instead of the misleading "has no content" error (which remains for genuinely content-less pastes).

Full suite green on both sides; all five CLI targets compile.

guiyanakuang:
Pushed 0ef06493: reverted the CLI API version bump from the previous commit — the CLI binary has never shipped in a release, so there is no deployed v1 peer to be compatible with; `CLI_API_VERSION` stays at 1 (with a comment noting no bump is needed until the CLI first ships). The defensive `--raw` error message for a server that returns no `rawContent` is kept — it still gives an actionable hint for mixed dev builds. Tests green on both sides.

Issue #4793, opened 2026-08-18
Windows: single-consumer clipboard event pipeline with burst coalescing and event-time source sampling


guiyanakuang:
## Background

Snipping Tool writes the clipboard twice per capture, 30–90 ms apart (#4737). The Windows message loop serializes `onChange()`, so the two `getContents()` calls do not run concurrently — but each one is an eager full snapshot (open clipboard → copy all convertible formats → close), and reads are retried with fresh snapshots when a read fails or returns invalid content (`ControlUtils.blockExponentialBackoffUntilValid`). Two back-to-back expensive snapshots land exactly in the window where Snipping Tool's save pipeline — which appears to retry nothing and give up silently — needs clipboard access.

The subsequent `consume()` coroutines *do* run concurrently, degrading post-processing (`updatePasteData` 22 ms → 135 ms in the #4737 logs), but they operate on already-taken snapshots and do not hold the native clipboard; they are **not** the direct cause of the save failures.

## Design: single-consumer event pipeline

Rework the collection entry in `WindowsPasteboardService`:

1. **Callback does minimal work**: record the clipboard sequence number, sample the source app **at event time** (binding source to sequence), signal a `Channel.CONFLATED`, and return immediately — no clipboard read on the message thread.
2. **A dedicated worker**:
   - waits for a short stability window; consecutive sequence updates within the window coalesce into a single read (window extension bounded);
   - takes **one** `getContents()` snapshot for the whole burst;
   - re-validates the sequence number after the snapshot; if it changed, discards the snapshot and reads the latest content — with a **bounded** retry count. On reaching the bound, accept the current snapshot (note: it may capture an intermediate state), still process the latest conflated event afterwards, and record a degradation metric;
   - consumes **serially** — no overlapping consume passes.
3. **Source attribution**: after coalescing, use the source bound to the sequence that was actually captured. Do not re-query the foreground app after the stability window — by then the foreground may have changed.
4. **Adaptive delay**: a normal single copy sees a minimal delay; detecting a second update within a short window extends the wait (external experience from Ditto suggests ~500 ms is effective). Final parameters must be determined empirically with **50+ consecutive capture runs** — single-run results are meaningless for a race.

## Positioning

Moving the read off the message thread is a threading-model improvement; the actual mitigation is the delay plus the reduction in full-snapshot count (a blocked message pump does not block Snipping Tool's own clipboard calls).

[...]

Issue #4791, opened 2026-08-18
Short-window keep-first dedup for duplicate image records from a single clipboard operation


guiyanakuang:
## Background

On Windows, Snipping Tool writes the clipboard twice per capture (30–90 ms apart; see #4737). Each clipboard event is collected independently, producing two identical history records and two identical PNGs. This duplication is **not caused by consumer concurrency** — even fully serialized collection would produce two records — and it persists because image records intentionally bypass the generic same-hash cleanup.

## Why the existing gate must stay

The `isRefFiles()` gate in `PasteReleaseService.releaseLocalPasteData` was added deliberately by #2693 as a resource-lifecycle guard:

- `PasteItem.clear()` deletes the real backing files for non-ref (`basePath == null`) file/image items;
- `markDeleteSameHash` schedules an immediate `DELETE_PASTE_TASK` (no delay);
- the task executor runs up to 10 tasks concurrently, so an older record's sync task may still be reading the very file being deleted.

Lifting the gate would reintroduce premature file deletion and sync failures. **Do not lift it.**

## Scope

This issue is a **fallback**, contingent on #4793: implement only if duplicate records still occur after the Windows event-coalescing pipeline lands and is validated with real capture runs.

## Design: short-window keep-first dedup

- When a locally collected image paste computes its final hash and a local image record with the same hash already exists within a short window (a few seconds), keep the **first** record and discard the **second**.
- Discarding the second record must be correctness-aware: at that point its DB row still holds the pre-collect placeholder item **without final file paths**, so simply calling `markDeletePasteData` would orphan the just-written PNGs. Either:
  - persist the final resource paths with a DELETED state before scheduling the delete task, or
  - use a dedicated discard API that clears resources from the in-memory final `PasteItem`.
- Create only a delete task for the second record — **never sync or rendering tasks**. The discarded record therefore never has in-flight consumers, which is what keeps this safe while leaving the #2693 guard intact.
- Do not reuse the existing 1-day `markDeleteSameHash` path.

## Rejected alternative

Keep-newest (deleting the first record) is rejected: the first record's sync task may already be reading its PNG; making that safe would require task-completion barriers, resource leases, or reference counting — no longer a low-risk fix.

## Precondition to verify

Image hashes are content-based (file names are excluded), so identical pixels through the same encoding path normally produce identical hashes — but confirm with real DB hashes from a reproduced Snipping Tool duplicate before implementing.

[...]

Issue #4737, opened 2026-08-04
Windows 11 25H2: Clipboard monitoring prevents Snipping Tool from automatically saving screenshots taken with PrtScr


Wu-Kuan-Yee:
### How frequently does the bug occur?

Always

### Description

# Title

Windows 11 25H2: Clipboard monitoring prevents Snipping Tool from automatically saving screenshots taken with PrtScr

## Environment

- **OS:** Windows 11 Pro for Workstations 25H2
- **OS Build:** 26200.8973
- **Windows Feature Experience Pack:** 1000.26100.344.0
- **CrossPaste Version:** 2.1.6 (2390)

## Description

When CrossPaste is running with **Clipboard Monitoring** enabled, screenshots taken with the **PrtScr** key are copied to the clipboard as expected, but they are **not automatically saved** to the default `Pictures\Screenshots` folder.

The issue disappears immediately after disabling **Clipboard Monitoring** (while CrossPaste continues running), and reappears immediately after enabling it again.

The issue is consistently reproducible.

## Steps to Reproduce

1. Launch CrossPaste.
2. Ensure **Clipboard Monitoring** is enabled.
3. Press **PrtScr**.
4. Capture an area using Windows Snipping Tool.
5. Check the `Pictures\Screenshots` folder.

## Expected Behavior

The screenshot should:

- be copied to the clipboard.
- be automatically saved to `Pictures\Screenshots`.

## Actual Behavior

The screenshot is copied to the clipboard, but **no image file is automatically saved**.

## Additional Test

I verified the issue by toggling Clipboard Monitoring while CrossPaste remained running:

| Clipboard Monitoring | Result |
| -------------------- | ------ |
| Enabled | Screenshot is copied to the clipboard, but **not automatically saved** |
| Disabled | Screenshot is copied to the clipboard **and automatically saved** |
| Enabled again | The problem immediately returns |

This indicates that the issue is specifically related to CrossPaste's clipboard monitoring functionality, rather than CrossPaste simply running in the background.

## Possible Cause

The issue appears to be related to CrossPaste's Windows clipboard monitoring implementation.

Clipboard monitoring may be reading, holding, converting, or rewriting image clipboard data in a way that interferes with the Windows 11 25H2 Snipping Tool automatic-save workflow.

If there is any debug build or logging that could help diagnose the issue, I'd be happy to help test it.

### Stacktrace & log output

```shell

```

### Can you reproduce the bug?

Always

### Reproduction Steps

[...]

PR #4798, opened 2026-08-18
:bug: discard duplicate local image records via short-window keep-first dedup

guiyanakuang:
## Background

On Windows, Snipping Tool writes the clipboard twice per capture (30–90 ms apart, #4737). Real-machine validation after the single-consumer event pipeline (#4793) confirmed duplicate image records still occur, so this implements the keep-first fallback designed in #4791.

Non-ref image records intentionally bypass the generic `markDeleteSameHash` cleanup (the #2693 resource-lifecycle guard), so these duplicates persist. That guard is left fully intact.

## Change

- New `getRecentSameHashLocalPasteId` query/DAO method (shared): probes for a LOADED, locally collected (`remote = 0`) record with the same hash and paste type inside an explicit `(minCreateTime, maxCreateTime)` range. The dedup window policy lives in `PasteReleaseService` (`IMAGE_DEDUP_WINDOW`, 5s), not in the DAO.
- `PasteReleaseService.releaseLocalPasteData`: locally collected non-ref image releases go through `releaseLocalImageDeduped`, which runs the duplicate probe and the row's final state commit (LOADED or DELETED) under a per-hash `StripedMutex` — without it, two concurrent releases of the same image could each miss the other's not-yet-committed record and both survive (macOS/Linux launch a consumer per clipboard change; only the Windows pipeline serializes consumes). Image IO and hashing happen earlier in collection and are never under the lock; non-image types don't touch the mutex.
- The window compares record createTimes (event proximity), bounded on both sides (`createTime ± window`): slow image processing before release cannot shrink it, and out-of-order completion cannot make a delayed release discard itself against an identical image copied long after its own event.
- Discarding the duplicate (keep-first; keep-newest was rejected in #4791 because the first record's sync task may already be reading its files):
  1. The final items (with the real file paths) are persisted into the row first — at this point the row still holds the pre-collect placeholder, and deleting it as-is would orphan the just-written PNGs. `writeLoadedRow` is the single writer of that statement, shared with the normal release path so the two branches cannot drift.
  2. The row is marked deleted in the same transaction.
  3. Only a `DELETE_PASTE_TASK` is created — never sync or rendering tasks — so the discarded record never has in-flight consumers and clearing its files cannot race anything.

No additional hashing is introduced: the content hash is already computed unconditionally during collection (`DesktopImageTypePlugin` → `getFileInfoTree` → streaming MurmurHash3), off the UI thread. The dedup adds one indexed query.

If the two writes of a capture ever produce different bytes, the hashes differ and the dedup is a no-op — fail-safe, no false deletions.

## Tests

`PasteReleaseServiceImageDedupTest` (real in-memory database + real `SqlPasteDao`), 9 cases:
- duplicate local image within the window is discarded, final paths persisted, only a delete task created, first record untouched, `currentPaste` not updated
- identical image outside the window is kept and released normally
- the window is measured between record createTimes, not release time (release delayed ~8s, records 50 ms apart → still deduped)
- a LOADED candidate newer than the window does not trigger dedup (out-of-order completion)
- two concurrent releases of the same image keep exactly one record
- a remote record with the same hash does not trigger dedup
- a LOADING row carrying the final hash is not a kept candidate
- the dedup query ignores deleted records
- text pastes keep the existing keep-new same-hash cleanup semantics

Note for mobile follow-up: `shared` changes (`PasteDatabase.sq` + `PasteDao`/`SqlPasteDao`).

Addresses #4791. Addresses #4737.

### DIFF
--- a/app/src/commonMain/kotlin/com/crosspaste/paste/PasteReleaseService.kt
+++ b/app/src/commonMain/kotlin/com/crosspaste/paste/PasteReleaseService.kt
@@ -131,6 +131,10 @@ class PasteReleaseService(
             val hash = firstItem.hash
             val pasteType = firstItem.getPasteType()
 
+            if (discardDuplicateLocalImage(pasteData, firstItem, remainingItems, size, hash, pasteType, id)) {
+                return@let
+            }
+
             val change =
                 database.transactionWithResult {
                     database.pasteDatabaseQueries.updatePasteDataToLoaded(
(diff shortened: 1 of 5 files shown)


guiyanakuang:
Applied three review fixes in 07f35fdc:

1. **Atomic probe + commit (P1)**: the duplicate probe and the row's final state commit (LOADED or DELETED) now run under a per-hash `StripedMutex` (`releaseLocalImageDeduped`). Without it, two concurrent releases of the same image could each miss the other's not-yet-committed record and both survive — macOS/Linux launch a consumer per clipboard change, so commonMain cannot rely on the Windows pipeline's serialized consumes. The lock covers only the query and the persistence decision for local non-ref images; image IO and hashing happen earlier in collection and are never under the lock, and non-image types don't touch the mutex at all.
2. **Window compares record createTimes (P2)**: `minCreateTime` is now derived from the releasing record's own `createTime` (set at pre-record creation, i.e. event time) instead of `now` at release time, so slow image encoding/hashing can no longer shrink the effective window. The `IMAGE_DEDUP_WINDOW` policy moved from `SqlPasteDao` to `PasteReleaseService`; the DAO takes an explicit `minCreateTime`.
3. **Candidate must be LOADED (P3)**: the query now requires `pasteState = 1` instead of `pasteState != -1`, no longer relying on the "LOADING rows have an empty placeholder hash" convention. `IdxPasteDataStateCreateTime(pasteState, createTime)` covers the new predicate.

New tests: a LOADING row carrying the final hash is not a kept candidate; the window is measured between record createTimes (release delayed ~8s, records 50ms apart → still deduped); two concurrent releases of the same image keep exactly one record. 8/8 passing, full app + shared suites green.

guiyanakuang:
Applied the remaining review points in 70a7ca71:

1. **Bounded window on both sides**: the probe now takes an explicit `(minCreateTime, maxCreateTime)` range of `createTime ± IMAGE_DEDUP_WINDOW`. Without the upper bound, out-of-order completion could make a delayed release discard itself against an identical image copied long after its own event (e.g. old release stuck 12s, same image re-copied 10s later and released first). New test: a LOADED candidate newer than the window does not trigger dedup.
2. **Shared loaded-row writer**: extracted `writeLoadedRow(...)` as the single writer of the `updatePasteDataToLoaded` statement; `persistLoaded` and the duplicate-discard transaction both call it with `pasteSearchContent` as the only differing argument, so the two branches cannot drift when fields are added.

9/9 dedup tests passing, full app + shared suites green.

PR #4803, opened 2026-08-18
:construction_worker: split the CLI native build into a parallel path-gated CI job

### LINKED ISSUE #4802: Speed up PR CI: run the CLI native build in a parallel job and only when CLI-relevant paths change
## Problem

After the test tiering in #4800, the PR `app-build` job takes ~7 minutes, of which:

- ~3.4 min: compile core + shared + app (+ ktlint + fast-tier tests)
- ~2.9 min: Kotlin/Native CLI build (`:cli:ktlintCheck :cli:cliNativeTest` + 3 link targets)

Two inefficiencies:

1. The CLI native build runs **serially after** the app build inside the same job, although the two share no build outputs.
2. It runs on **every** app PR, although the `cli` module has no project dependencies (it is a pure Kotlin/Native HTTP client since the all-HTTP refactor) — changes under `app/`, `shared/`, or `core/` cannot affect it.

## Proposal

- Split the CLI native build into its own `cli-build` job so it runs in parallel with `app-build`.
- Add a `cli` paths filter (the `changes` job already exists): only `cli/**` and shared Gradle build configuration trigger it. Remove `cli/**` from the `app` filter so CLI-only changes no longer trigger the app build either.
- Include `cli-build` in the final `ci` aggregation gate (skipped counts as pass, failure/cancelled fails, same as the existing jobs).

Expected effect: a typical app-only PR drops from ~7 min to ~4 min (no K/N CLI build at all); a CLI-touching PR stays ~7 min worth of compute but finishes in ~4 min wall clock thanks to parallelism. Release workflows are unaffected — they keep building and linking the CLI unconditionally.

guiyanakuang:
## Change

Implements both optimizations from #4802:

1. **Parallel job**: the CLI native build (`:cli:ktlintCheck :cli:cliNativeTest` + linuxX64/linuxArm64/mingwX64 links) moves out of `app-build` into its own `cli-build` job with its own Konan toolchain cache and CI Gradle overrides, so it runs concurrently with the app build.
2. **Path gating**: a new `cli` filter in the existing `changes` job triggers `cli-build` only for `cli/**` and shared Gradle build configuration. The `cli` module has no project dependencies (pure Kotlin/Native HTTP client), so `core`/`shared`/`app` changes cannot affect it. `cli/**` is likewise removed from the `app` filter, so CLI-only changes no longer rebuild the desktop app.
3. The final `ci` aggregation gate now includes `cli-build` with the same semantics as the other jobs: skipped counts as pass, failure/cancelled fails the gate.

Release workflows are unaffected: `build-release.yml` and `build-beta-linux.yml` keep building and linking the CLI unconditionally.

## Expected effect

| PR touches | Before | After |
|---|---|---|
| app/shared/core only | ~7 min (serial K/N CLI build included) | **~4 min** (cli-build skipped) |
| cli only | ~7 min | **~3 min** (app-build skipped) |
| both | ~7 min | ~4 min wall clock (parallel) |

YAML validated structurally (jobs/needs/outputs/if wiring). This PR itself only touches `.github/workflows/ci.yml`, so this run exercises the `changes` filter fallthrough (both `app` and `cli` trigger on workflow changes) and the parallel execution path.

Fixes #4802.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,12 +93,18 @@ jobs:
 
       # The Kotlin/Native toolchain dependencies (gcc sysroot, llvm, lldb, libffi)
       # are downloaded and extracted into ~/.konan on every run otherwise.
+      # Job-scoped key: app-build and cli-build need different toolchain pieces
+      # (cli cross-compiles linuxArm64/mingwX64), and actions/cache never
+      # re-saves on an exact-key hit — a shared key seeded by one job would
+      # permanently miss the other job's dependencies and re-download them on
+      # every run. The bare konan- fallback bootstraps from pre-split caches.
       - name: Cache Kotlin/Native toolchain
         uses: actions/cache@v5
         with:
           path: ~/.konan
-          key: konan-${{ runner.os }}-${{ hashFiles('gradle/libs.versions.toml', 'gradle/wrapper/gradle-wrapper.properties') }}
+          key: konan-app-${{ runner.os }}-${{ hashFiles('gradle/libs.versions.toml', 'gradle/wrapper/gradle-wrapper.properties') }}
           restore-keys: |
+            konan-app-${{ runner.os }}-
             konan-${{ runner.os }}-
 
       - name: Install native build dependencies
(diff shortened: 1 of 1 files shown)


guiyanakuang:
Pushed a second commit (837cfc34) fixing why the existing `~/.konan` cache never stopped the cross-target downloads seen in earlier logs (aarch64 gcc sysroot, qemu, msys2-mingw on every run):

`actions/cache` never re-saves on an exact-key hit. Once the shared `konan-<os>-<hash>` key was seeded by a run that lacked the cross-compilation dependencies, every later run hit that incomplete cache and re-downloaded the missing pieces forever. Splitting `cli-build` out would have made this structural: two jobs needing different toolchain pieces racing for one key.

The cache key is now job-scoped (`konan-app-...` / `konan-cli-...`), so each job's first save contains exactly the dependency set that job needs and subsequent exact hits are complete. A bare `konan-<os>-` restore fallback bootstraps both from the pre-split caches.

guiyanakuang:
Measured on this PR's own run (32137550837), which touches `ci.yml` and therefore triggers every filter — the worst case:

| Job | Duration |
|---|---|
| app-build (CLI step removed) | 4m26s (was ~7m serial) |
| cli-build (parallel; cold cross-toolchain downloads, now saved under `konan-cli-*`) | 4m19s |
| web-build | 1m55s |
| **Wall clock** | **~4.5 min** (was ~7 min) |

A typical app-only PR skips `cli-build` entirely (~4.5 min total); a CLI-only PR skips `app-build` (~4.3 min). This run seeded the job-scoped `konan-cli` cache including the linuxArm64/mingwX64 sysroots and qemu, so subsequent CLI runs should not re-download them.

PR #4804, opened 2026-08-18
:bug: Recover persisted tasks after restart and enforce task-aware shutdown ordering

### LINKED ISSUE #4797: Recover persisted tasks after restart and enforce task-aware shutdown ordering
## Problem

`PasteTask` records are persisted in the database, but task scheduling is only maintained in the in-memory `Channel<Long>` owned by `TaskExecutor`.

A task can therefore become permanently orphaned in either of these cases:

1. The task record is committed, but the process exits or crashes before its ID is submitted to `TaskExecutor`.
2. A task producer is still draining during application shutdown after `TaskExecutor.shutdown()` has closed the channel.
3. Execution is cancelled after the task status has changed from `PREPARING` to `EXECUTING`.

On the next application start, `TaskExecutor` only starts consuming its new in-memory channel. It does not query the database for unfinished tasks. The task DAO currently supports lookup by task ID but has no query for resumable `PREPARING` or stale `EXECUTING` tasks.

As a result, persistence records the unfinished work but does not cause that work to resume. These records can remain in the database indefinitely, and the corresponding synchronization, deletion, cleanup, or metadata operation may never complete.

## Context

This was identified while reviewing the Windows clipboard single-consumer pipeline and application shutdown behavior. The underlying recovery gap and concurrent service shutdown behavior already exist on the base branch; they were not introduced by the clipboard pipeline change. This issue should be addressed independently.

## Expected behavior

Once a task has been durably recorded, it should either:

- complete during the current process lifetime, or
- be discovered and safely resumed after the next application start.

Application shutdown should also stop task producers before stopping the task executor, so normal shutdown does not unnecessarily rely on crash recovery.

## Proposed direction

1. Add DAO queries for resumable `PREPARING` tasks and stale `EXECUTING` tasks.
2. Define a stale policy for `EXECUTING` tasks left by an interrupted process.
3. Atomically reset or claim recoverable tasks before placing their IDs into the new `TaskExecutor` channel.
4. Run recovery during application startup after task executors are registered and before new producers begin submitting work.
5. Review every persisted task type for retry idempotency and duplicate side effects.
6. Make shutdown dependency-aware: stop and drain task producers such as the pasteboard and schedulers before shutting down `TaskExecutor`.
7. Keep recovery bounded and observable, with logs for recovered, skipped, and failed tasks.

## Acceptance criteria

- A task committed as `PREPARING` but never submitted is executed after restart.
- A stale `EXECUTING` task follows the documented recovery policy after restart.
- `SUCCESS` and terminal `FAILURE` tasks are not replayed.
- Recovery cannot enqueue the same task more than once during concurrent startup.
- Retried task types do not produce unsafe duplicate side effects.
- Normal application shutdown stops task producers before `TaskExecutor`.
- Tests cover interruption between database commit and queue submission, interruption during execution, and a clean shutdown with an in-flight producer.

guiyanakuang:
Fixes #4797

## Problem

`PasteTask` rows are durable, but scheduling only lived in `TaskExecutor`'s in-memory channel. A task became permanently orphaned when:

1. the row was committed but the process died before the ID was submitted to the channel,
2. a producer was still draining during shutdown after the channel had been closed (`send` threw), or
3. execution was cancelled after the status had moved from `PREPARING` to `EXECUTING`.

On the next start nothing queried the database for unfinished work, so those rows sat there forever and the corresponding sync/delete/cleanup never completed.

### Startup recovery

- **New DAO queries `getMaxTaskId()` + `claimRecoverableTasks(maxTaskId, limit = 1000)`** (`PasteTaskDatabase.sq`, `TaskDao`, `SqlTaskDao`): one transaction selects up to `limit` IDs of `PREPARING`/`EXECUTING` rows with `taskId <= maxTaskId` and resets the claimed `EXECUTING` ones back to `PREPARING`. Overflow rows keep their status and are claimed on a later startup, keeping recovery bounded.
  - **Stale policy**: the single-instance app lock guarantees that at startup every unfinished row at or below the bound belongs to a dead process, so all of them are recoverable. The bound is a **task-id watermark**, not wall-clock time: ids are monotonic (AUTOINCREMENT), so the guarantee is immune to system clock changes.
- **`TaskExecutor.recoverPersistedTasks(maxTaskId)`**: claims and re-enqueues, guarded to run at most once, logs the recovered count plus a capped id sample (or that there was nothing to recover) and any claim failure.
- **Startup ordering** (`CrossPaste.startApplication`): the watermark is captured **before `SyncManager.start()`**, i.e. before any task producer of the current process is live. Any task a producer creates afterwards has a larger id and can never be claimed, so double-enqueue is impossible by construction rather than by timing. Recovery itself runs right after `SyncManager.start()` / `PastePullService.init()` and before the pasteboard listener, servers, CLI, MCP and `CleanScheduler` start.
- **`GeneralSyncManager.start()` now waits (bounded, 3s) for the first device snapshot** before returning: the DB-flow collector that builds sync handlers is asynchronous, and recovery runs immediately after `start()`. Without the gate, recovered `SYNC_PASTE` tasks would see an empty handler map and complete without syncing, and recovered `PULL_FILE` tasks could burn their retries instantly (retry exhaustion deletes the incomplete paste). The wait is a local DB read (normally milliseconds); the timeout only guards a stuck flow and logs a warning.

### Task-aware shutdown

- **`TaskExecutor.shutdown()`** is now suspend: it closes the channel, drains queued and in-flight tasks for up to 2s, then cancels the scope **in a `finally`** — even when an enclosing shutdown timeout cancels the drain, task coroutines cannot outlive the closed database. Tasks still running after the drain keep their persisted status and are resumed on the next start.
- **`submitTask` uses `trySend`**: a producer racing shutdown no longer gets an exception; the row stays `PREPARING` and is recovered on the next start (with a warning log).
- **`executeTask` rethrows `CancellationException`** instead of recording a cancelled execution as terminal `FAILURE` — the row keeps its status for recovery.
- **`executeTask` failure recording is itself fault-tolerant**: `createFailExtraInfo` parses the persisted extraInfo, and a corrupt row used to throw out of the error path — cancelling the consumer coroutine and permanently stopping all task processing (and recovery would re-feed the row every start). It now falls back to recording the failure without new extraInfo, and any persistence error is logged instead of propagating.
- **Unknown task types are left untouched**: a downgraded build that sees a task type persisted by a newer version logs a warning and leaves the row `PREPARING` (previously it was flipped to `EXECUTING` and then terminal `FAILURE`), so a later upgrade can still recover it.
- **`shutdownAllServices()` is now three-phased** inside the same 5s budget: (1) stop all task producers (pasteboard, scheduler, servers, GUI services) in parallel, (2) drain `TaskExecutor`, (3) stop the services drained tasks still depend on (`SyncManager`, rendering, HTTP clients) in parallel, then close the driver. Phase 1 has its own 2s sub-budget so one slow stop (e.g. a server's grace period) cannot starve the drain and phase 3; a producer that outlives the budget can only `trySend` into a closed channel, which defers its task to recovery. `stopManagedService` rethrows `CancellationException` so an expired phase budget actually skips ahead instead of being swallowed, and the Bonjour teardown (which blocks in its own `runBlocking` for up to 4s and is not a producer) runs concurrently with all phases instead of eating a phase budget.

## Retry-idempotency review of persisted task types (acceptance item 5)

All task types already support retry (`needRetry=true` re-submits), so a restart replay is no stronger than the existing retry path:

- `SYNC_PASTE_TASK`: a restart replay has the same duplicate exposure as the existing retry path (a push whose response was lost is already retried today), and the retry ceiling (3 execution histories) still applies after recovery. The snapshot gate in `GeneralSyncManager.start()` ensures recovered tasks see the known handlers.
- `DELETE_PASTE_TASK` / `CLEAN_PASTE_TASK` / `CLEAN_TASK_TASK` / `PULL_ICON_TASK` / `OPEN_GRAPH_TASK`: idempotent (delete/cleanup of missing rows is a no-op, downloads overwrite).
- `PULL_FILE_TASK`: `PullExtraInfo` tracks progress; re-pull overwrites chunks.
- `SWITCH_LANGUAGE_TASK`: guarded by a current-language check; replay of an outdated switch is skipped.
- `DELAYED_DELETE_PASTE_TASK`: `executeAfter` is an absolute timestamp, so a recovered task re-schedules correctly (immediately if past due). Pre-existing limitation (unchanged): the task marks `SUCCESS` before the in-memory delayed delete fires, so a crash inside that window still loses the delete — tracked as a possible follow-up, not a regression of this PR.

## Acceptance criteria mapping

- Committed-but-never-submitted `PREPARING` task executes after restart → `TaskRecoveryTest`.
- Stale `EXECUTING` task follows the documented policy (reset + re-run) → `TaskRecoveryTest`, `TaskDaoTest`.
- `SUCCESS`/terminal `FAILURE` never replayed → `TaskDaoTest`.
- No double-enqueue during concurrent startup → single-run guard + claim-before-producers ordering (`TaskExecutorTest`).
- Producers stop before `TaskExecutor` on normal shutdown → phased `shutdownAllServices()`.
- Clean shutdown with an in-flight producer → `trySend` path test (`TaskRecoveryTest`, `TaskExecutorTest`).

## Known limitation (documented, unchanged behavior)

The `CancellationException` rethrow in `TaskExecutor.executeTask` protects the framework path (DB writes around execution), but individual `SingleTypeTaskExecutor` implementations that wrap their network calls in `runCatching` still swallow cancellation internally and convert it into their normal fail/retry path. That is pre-existing behavior; tightening every executor is out of scope here.

## Mobile follow-up

`shared` changed (`PasteTaskDatabase.sq` + `TaskDao`/`SqlTaskDao`: `getMaxTaskId`, `claimRecoverableTasks`) and `app/commonMain` changed (`TaskExecutor.shutdown()` is now suspend, new `recoverPersistedTasks`, `GeneralSyncManager.start()` gates on the first snapshot): mobile repos need these when syncing shared code and should wire an equivalent startup recovery call and shutdown ordering.

## Testing

- `./gradlew app:desktopTest` — full fast tier green; new suites stay in the fast tier (in-memory SQLite, virtual time; `TaskRecoveryTest` uses real dispatchers but completes in milliseconds).
- New regression coverage: bounded claim with overflow left intact, id-watermark exclusion, unknown-task-type skip, corrupt-extraInfo consumer survival, drain-then-cancel shutdown, submit-after-shutdown, and `GeneralSyncManager.start()` waiting for a delayed first snapshot (plus its bounded-timeout path).

### DIFF
--- a/app/src/commonMain/kotlin/com/crosspaste/sync/GeneralSyncManager.kt
+++ b/app/src/commonMain/kotlin/com/crosspaste/sync/GeneralSyncManager.kt
@@ -20,6 +20,7 @@ import com.crosspaste.utils.namedScope
 import io.github.oshai.kotlinlogging.KotlinLogging
 import io.ktor.util.collections.*
 import kotlinx.atomicfu.atomic
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
@@ -37,7 +38,9 @@ import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import kotlinx.coroutines.withTimeoutOrNull
 import kotlin.coroutines.cancellation.CancellationException
+import kotlin.time.Duration.Companion.seconds
 
 class GeneralSyncManager(
     override val realTimeSyncScope: CoroutineScope = namedScope(ioDispatcher, "GeneralSyncManager"),
@@ -87,6 +90,14 @@ class GeneralSyncManager(
 
     private val started = atomic(false)
 
+    /**
+     * Completed after the first device snapshot from the database has been
+     * turned into sync handlers. [start] waits for it (bounded) so callers —
+     * in particular startup task recovery, which runs right after [start]
+     * returns — never observe an empty handler map for known devices.
+     */
+    private val initialSnapshotLoaded = CompletableDeferred<Unit>()
+
     private var syncRuntimeInfosJob: Job? = null
 
     override suspend fun start() {
@@ -122,6 +133,11 @@ class GeneralSyncManager(
         }
 
         startCollectingSyncRuntimeInfosFlow()
+
+        // Bounded: the first emission is a local DB read that normally lands in
+        // milliseconds; the timeout only guards startup against a stuck flow.
+        withTimeoutOrNull(3.seconds) { initialSnapshotLoaded.await() }
+            ?: logger.warn { "Timed out waiting for the initial device snapshot" }
     }
 
     override suspend fun stop() {
@@ -186,6 +202,8 @@ class GeneralSyncManager(
                             .filter {
                                 it in currentAppInstanceIdSet
                             }.toSet()
+
+                    initialSnapshotLoaded.complete(Unit)
                 }
             }
     }
--- a/app/src/commonMain/kotlin/com/crosspaste/task/TaskExecutor.kt
+++ b/app/src/commonMain/kotlin/com/crosspaste/task/TaskExecutor.kt
@@ -2,21 +2,28 @@ package com.crosspaste.task
 
 import com.crosspaste.db.task.PasteTask
 import com.crosspaste.db.task.TaskDao
+import com.crosspaste.utils.DateUtils.nowEpochMilliseconds
 import com.crosspaste.utils.TaskUtils
 import com.crosspaste.utils.cpuDispatcher
 import com.crosspaste.utils.namedScope
 import io.github.oshai.kotlinlogging.KotlinLogging
+import kotlinx.atomicfu.atomic
 import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Semaphore
+import kotlinx.coroutines.withTimeoutOrNull
+import kotlin.coroutines.cancellation.CancellationException
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
 
 class TaskExecutor(
     singleTypeTaskExecutors: List<SingleTypeTaskExecutor>,
     private val taskDao: TaskDao,
     maxConcurrentTasks: Int = 10,
+    private val drainTimeout: Duration = 2.seconds,
     private val scope: CoroutineScope = namedScope(cpuDispatcher, "TaskExecutor"),
 ) {
     private val logger = KotlinLogging.logger {}
@@ -27,7 +34,9 @@ class TaskExecutor(
 
     private val semaphore = Semaphore(maxConcurrentTasks)
 
-    init {
+    private val recoveryStarted = atomic(false)
+
+    private val consumerJob =
         scope.launch(CoroutineName("TaskExecutor")) {
             for (taskId in taskChannel) {
                 semaphore.acquire()
@@ -40,7 +49,6 @@ class TaskExecutor(
                 }
             }
         }
-    }
 
     private fun getExecutorImpl(taskType: Int): SingleTypeTaskExecutor =
         singleTypeTaskExecutorMap[taskType] ?: throw IllegalArgumentException("Unknown task type: $taskType")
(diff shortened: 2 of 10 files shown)


guiyanakuang:
Closing after re-evaluating cost/benefit. The orphan windows this addresses are rare (millisecond-wide crash windows, shutdown races), and most task types are already self-healing at the product level: periodic cleanup tasks are re-created every cycle, a missed clipboard sync is repaired by the next copy, and previews/icons are cosmetic. Meanwhile the recovery machinery touches the most sensitive lifecycle code (startup ordering, shutdown orchestration, SyncManager gating), carries a mobile follow-up obligation for the shared/commonMain changes, and replaying a stale SYNC_PASTE task at startup would surprisingly push old clipboard content to peers — for ephemeral clipboard data the existing at-most-once semantics are arguably correct.

Three genuine standalone defects found during this work (a corrupt extraInfo row permanently killing the task consumer, cancellation being recorded as terminal FAILURE, and producers crashing on the closed channel during shutdown) will be extracted into a separate small PR.

PR #4809, opened 2026-08-19
:sparkles: fit CLI list previews to the terminal width with cell-accurate truncation

### LINKED ISSUE #4807: CLI history/search previews are cut at a fixed 60 chars and misalign with CJK or multi-line content
## Current behavior

The table output of `crosspaste history` and `crosspaste search` truncates every preview at a hard-coded 60 chars (`preview.take(60)`), regardless of how wide the terminal window is. Two further problems compound it:

- Truncation counts chars, not display cells, so CJK/emoji previews (2 cells per glyph) look inconsistent next to ASCII rows and can overflow the intended budget.
- Only `\n` is replaced before printing; other whitespace runs (tabs, leading indentation from HTML-derived summaries) survive and visually misalign the preview column.

## Expected behavior

The preview column should fill whatever terminal width is left after the fixed columns, truncate by display cells so wide glyphs never make the row wrap, and collapse embedded whitespace runs so rows stay aligned.

Non-interactive stdout (pipes, Gradle-forwarded output) cannot be probed for a window size; an explicit `COLUMNS` environment variable should be honored there before falling back to the default width.

## Notes

The server-side summary (`PasteDataHelper.getSummary`) already returns the full text; this is purely a CLI rendering concern. Mordant's table widget was evaluated and rejected: its cell truncation (`Span.take`) counts chars while its budget is in cells (mordant 3.0.2), so a NOWRAP cell full of wide chars renders wider than the column.

guiyanakuang:
Closes #4807

## Summary

- `history`/`search` table previews now fill whatever terminal width is left after the fixed columns instead of being cut at a hard-coded 60 chars.
- Truncation is done by display cells via an approximate wcwidth (wide CJK blocks and emoji count 2, joiners/variation selectors 0), deliberately biased toward overestimating: a preview truncated one cell early is invisible, one cell late wraps the whole row. Surrogate pairs are never split.
- Embedded newlines and whitespace runs collapse to single spaces, fixing the misaligned preview column for HTML-derived summaries.
- Non-interactive stdout (pipes, Gradle-forwarded output) cannot be probed for a window size, and mordant ignores `COLUMNS` there — the CLI now honors an explicit `COLUMNS` before accepting the 79-column fallback.
- Mordant's table widget was evaluated and rejected for this: its cell truncation (`Span.take`) counts chars while its budget is in cells (mordant 3.0.2), so a NOWRAP cell full of wide chars renders wider than the column and wraps.

## Test plan

- New `PasteListRenderingTest` (8 cases): ASCII within/over budget, CJK cell counting, mixed content, surrogate-pair emoji never split, VS16 no undercount, zero/negative budgets, whitespace collapsing.
- `./gradlew :cli:macosArm64Test` green; mingwX64/linuxX64 compile clean.
- Manually verified against a running app: ASCII, CJK, and emoji rows all stay on one line at the exact terminal width, `COLUMNS=140` widens piped output accordingly.


guiyanakuang:
Addressed review feedback on the width table: the hand-picked ranges underestimated several East_Asian_Width Wide chars outside the CJK blocks (⌚ U+231A, ⏩ U+23E9-23EC, Tangut, Nushu, ...), which could still wrap a row when such a char landed at the line end.

- The ranges are now generated from the authoritative `EastAsianWidth.txt` (Unicode 16.0.0): every W/F codepoint plus the blocks whose unassigned codepoints default to Wide, merged into a binary-searched interval table (`EastAsianWideRanges.kt`, 120 intervals). `scripts/generate-cli-cell-width-table.py` regenerates it against any future UCD version.
- Emoji_Presentation chars are a subset of W/F since Unicode 9, so no separate emoji table is needed; VS16 keeps its +1 overestimate bias.
- Truncation is now cluster-aware: a base char is kept together with its trailing zero-width marks/joiners and VS16, so truncation can no longer strip a variation selector and silently change the glyph's presentation.
- New tests: ⌚/⏩ cell width and end-of-line drop, non-BMP Tangut U+17000 (surrogate pair), all 10 green; mingwX64/linuxX64 compile clean.

PR #4836, opened 2026-08-21
:sparkles: show the command line extension entry on every platform and in dev runs

### LINKED ISSUE #4835: Show the command line extension entry on every platform and in dev runs
## Problem

Validating 2.2.0.2502 on Windows surfaced that the Extension page has no "Command Line" entry. This was by design (D6): `CliSymlinkService` reported `NOT_SUPPORTED` everywhere except macOS production installs with the bundled CLI, and the entry row was hidden in that state — so Windows, Linux, and every dev run showed nothing.

Decision: that part of D6 is reversed. Platform behavior should be close to identical — the entry must be visible on every desktop platform and in dev runs. The PATH wiring itself (deb symlink, MSIX execution alias, in-app symlink install on macOS) is unchanged.

## Changes

- Extension page: the "Command Line" row is always shown; the sub-page adapts per platform instead of hiding.
- `CliSymlinkState`: remove `NOT_SUPPORTED`; add `PROBING` (initial state before the startup probe), `BINARY_MISSING` (no CLI executable found), and `EXTERNALLY_MANAGED` (Windows/Linux — PATH is wired by the installer; the page shows the executable path and shell availability, plus PATH guidance when the command does not resolve).
- macOS keeps the full install/repair state machine (TRANSLOCATED/INSTALLED/NOT_INSTALLED/NEEDS_REPAIR/CONFLICT), now also in dev runs (installing links to the locally built CLI).
- Dev binary resolution: `cliBinaryPath` in `development.properties` wins when set — a configured-but-missing path fails fast at startup; otherwise the conventional `cli/dist/<hostTarget>/` output of `./gradlew :cli:assembleDist` is discovered by walking up from the working directory; when nothing is found the page shows a `BINARY_MISSING` hint instead of blocking unrelated app development.
- Shell availability probing goes cross-platform: `where.exe crosspaste-cli` (cmd view of PATH) plus Windows PowerShell 5.1-safe `Get-Command` on Windows; zsh/bash filtered by existence (with /bin/sh fallback) on Unix, since Linux often lacks zsh.
- The one-time `InstallCliDialog` prompt stays macOS-only.
- i18n: 4 new keys across all 10 locales; `install_cli_tool_desc` reworded to be command-name-agnostic (the Windows command is `crosspaste-cli`, not `crosspaste`).

guiyanakuang:
Closes #4835

## Summary

Reverses the settings-entry half of design decision D6: the Extension page "Command Line" entry was only rendered on macOS production installs (`CliSymlinkService` reported `NOT_SUPPORTED` everywhere else, hiding the row), so Windows/Linux users and every dev run saw nothing. The entry is now always visible and the sub-page adapts per platform. The PATH wiring itself (deb symlink, MSIX execution alias, in-app symlink install on macOS) is unchanged.

## Changes

- **State machine** (`CliSymlinkState`): `NOT_SUPPORTED` removed. New states: `PROBING` (initial, before the startup probe), `BINARY_MISSING` (no CLI executable found), `EXTERNALLY_MANAGED` (Windows/Linux — the installer owns PATH wiring; the app only observes). macOS keeps TRANSLOCATED/INSTALLED/NOT_INSTALLED/NEEDS_REPAIR/CONFLICT with install/repair, now also available in dev runs (linking to the locally built CLI, with the translocation/conflict guards unchanged).
- **Dev binary resolution**: `cliBinaryPath` in `development.properties` wins when set; a configured-but-missing path fails fast (`check`) at first use, consistent with DevConfig's existing fail-fast style. Unset, the conventional `cli/dist/<hostTarget>/crosspaste-cli[.exe]` output of `./gradlew :cli:assembleDist` is discovered by walking up from the working directory. Nothing found → `BINARY_MISSING` with a hint on the CLI page, never a startup crash — building the CLI stays opt-in for app development that does not touch it.
- **Shell availability probing** is now cross-platform: Windows probes `where.exe crosspaste-cli` (the cmd view of PATH, which covers the WindowsApps alias directory) and Windows PowerShell `Get-Command` (5.1-safe, plain cmdlet + string concatenation); Unix keeps the login-shell `command -v` probe, with zsh/bash filtered by existence and a /bin/sh fallback since Linux often lacks zsh. The non-blocking line-scanning reader is generalized to take a per-probe line parser; the final unterminated line is now also parsed on process exit (where.exe output robustness).
- **UI**: the Extension row is unconditional; `CliToolSettingItem` renders per state (install/repair buttons on macOS, installer-managed status text on Windows/Linux, missing-binary hint with the dev build command in dev runs); the executable-path section shows an explicit not-found message instead of `null`.
- The one-time `InstallCliDialog` startup prompt remains macOS-only, gated on `NOT_INSTALLED` as before.
- **i18n**: 4 new keys (`cli_binary_missing`, `cli_binary_missing_desc`, `cli_binary_missing_dev_desc`, `cli_tool_external_desc`) across all 10 locales; `install_cli_tool_desc` reworded to be command-name-agnostic — the Windows command is `crosspaste-cli` (the `crosspaste.exe` alias is Conveyor's GUI launcher), so "use the crosspaste command" was wrong there.

## Testing

- `CliSymlinkServiceTest` grown from 24 to 33 cases: initial `PROBING` state, `BINARY_MISSING` (absent binary, wrong-name binary on Windows), `EXTERNALLY_MANAGED` on Windows/Linux, install refusal in both new states, dev resolution (configured-valid / configured-missing-throws / convention walk-up / nothing-found-null), dist-target and binary-name mapping, `where.exe` output parsing.
- Full fast-tier suite (`./gradlew app:desktopTest`) passes; `ktlintFormat` clean.


guiyanakuang:
Review fixes (521ee6d8), three P2s:

1. **CLI built at runtime was never picked up** — `cliBinaryPath` was resolved once at construction, so a dev run started before `:cli:assembleDist` stayed `BINARY_MISSING` forever. The binary location is now re-resolved on every `refresh()` and exposed as a `StateFlow`; the packaged path and an explicitly configured `cliBinaryPath` (with its fail-fast validation) are still fixed at construction — only the cli/dist convention walk is re-evaluated.
2. **Relative configured path installed a dangling symlink and reported INSTALLED** — a relative `cliBinaryPath` was written verbatim into /usr/local/bin/crosspaste (resolving relative to /usr/local/bin), while the state comparison also compared the relative path and claimed success. Configured paths are now converted to absolute normalized form before validation and return; covered by a new test.
3. **AppImage users got invalid PATH guidance** — the EXTERNALLY_MANAGED subtitle told users to add the executable's folder to PATH, but AppImage mounts at a new temporary path every launch (documented D6 limitation). The runtime is now detected via the APPIMAGE environment variable and shows a dedicated message (new `cli_tool_appimage_desc` key, all 10 locales) pointing at the deb/tarball packages instead.

`CliSymlinkServiceTest` is at 35 cases; full fast tier green.

Issue #4848, opened 2026-08-21
CLI: wire sixel rendering into paste and pick, verify on Windows Terminal


guiyanakuang:
Phase 4 of #4844 — depends on #4845, #4846, and #4847.

Scope:

- `PasteImageOutput.kt` (`InlineImageRenderer`): add the SIXEL branch. Unlike ITERM/KITTY (which stream the stored file), sixel fetches exact-size RGBA from the app via `CliClient.getRawImage` and encodes with `writeSixelInlineImage`. Keep the existing resource ceilings, expressed as a cap on the requested pixel box; keep the fetcher injectable for tests. Consider a small `ImageEmitter` strategy interface so the renderer stops branching on protocol internals (each emitter declares whether it wants file bytes or RGBA) — but no broader abstraction than that.
- `pick/PickTui.kt` (`drawInlineImage`): sixel branch computes the pixel box from the fitted cell box x the probed cell pixel size, requests exact dimensions, positions the cursor and emits. No kitty-style explicit delete: full-frame repaint with spaces should erase the sixel region (verify on real WT). Reserve panel rows as `ceil(imageHeight / cellHeightPx)`.
- 404 from the endpoint (old app) → fall back to listing paths, same UX as today.

Fast-tier tests: renderer sixel path with a fake fetcher (RGBA response → expected output; 404 → fallback note), pick box computation with probed vs default cell metrics.

Real-machine verification (Parallels Windows VM):

- Windows Terminal >= 1.22: `paste` single/multiple images, `pick` preview panel, transparent PNG, JPEG.
- Degradation matrix: conhost (no `WT_SESSION`), `paste | cat` (stdout not a TTY), old app + new CLI (404 fallback), WT < 1.22 if obtainable (DA1 reply without attribute 4 → paths).
- `pick`: sixel region correctly erased by full repaints; scrolling behavior after `paste` output.

guiyanakuang:
Merged in #4852 and shipped in 2.2.0. Closing.

Issue #4844, opened 2026-08-21
Sixel graphics support in the CLI for Windows Terminal (tracking)


guiyanakuang:
## Motivation

Windows Terminal (>= 1.22) supports the sixel graphics protocol, and it is the default terminal for our Windows users — currently the CLI can only print file paths there. The CLI's inline image rendering (`paste` previews, `pick` preview panel) supports the iTerm OSC 1337 and kitty graphics protocols, both of which pass the encoded image file through for the terminal to decode. Sixel is fundamentally different: the sender must hold decoded pixels and perform palette quantization + encoding itself.

## Design summary

Full design: `ai/design/cli-sixel-support.md` (local design doc).

Key decisions:

1. **Pixels come from the app, not a CLI-side image decoder.** New endpoint `GET /cli/paste/{id}/image?index=&maxWidth=&maxHeight=` (desktopMain, `CliRouting.kt`) decodes and aspect-fit scales via the app's existing Skia stack and returns straight-alpha RGBA8888 with dimensions in response headers. `CLI_API_VERSION` 1 → 2; the CLI degrades to path fallback on 404 (old app). This keeps the "CLI never decodes pixels" doctrine, avoids adding C decoders (stb_image) to the K/Native build, and keeps untrusted image parsing out of the CLI process. Quantization + sixel emission in the CLI is pure math on raw pixels.
2. **Detection = env candidates + DA1 probe confirmation.** `WT_SESSION` / `TERM` containing `foot`/`mlterm`/`yaft` only nominate sixel as a candidate; a `CSI c` (DA1) probe must report attribute `4` before sixel is used, so an older Windows Terminal (< 1.22) falls back to paths instead of printing escape garbage. The same probe issues `CSI 16 t` to learn the cell pixel size (fallback assumption 10x20 px). Probing runs only for sixel candidates (zero cost for kitty/iTerm users) and only when stdin+stdout are both TTYs. Requires a new expect/actual `TerminalQuery` primitive (Mordant's RawModeScope parses key events and would eat the DA response).
3. **Encoder is pure Kotlin and fully unit-testable.** Median-cut adaptive palette (<= 256 colors), alpha < 128 → transparent (DCS `P2=1`), RLE emission, golden-string tests. No dithering in v1.
4. **Protocol priority stays KITTY > ITERM > SIXEL** — mintty/WezTerm keep using the higher-fidelity iTerm protocol; sixel serves Windows Terminal, foot, mlterm.

## Phases

[...]

PR #4850, opened 2026-08-21
:sparkles: add raw RGBA transcode endpoint for CLI image previews

### LINKED ISSUE #4846: App: raw RGBA transcode endpoint for CLI image previews
Phase 2 of #4844.

Add a desktopMain-only endpoint so the CLI can obtain decoded pixels for sixel rendering without ever decoding image files itself:

```
GET /cli/paste/{id}/image?index=0&maxWidth=<px>&maxHeight=<px>
→ 200
  X-Image-Width: <w>
  X-Image-Height: <h>
  Content-Type: application/octet-stream   (straight-alpha RGBA8888, row-major)
→ 404 when the paste has no stored image at that index
```

Scope:

- Route in `app/src/desktopMain/kotlin/com/crosspaste/cli/CliRouting.kt`, service logic beside `CliApi.kt`; decode + aspect-fit scale with the app's existing Skia stack (same infrastructure as `DesktopThumbnailLoader`). No commonMain changes, so no mobile impact.
- Address images by paste id + index — the server resolves its own stored paths; no path parameter (no arbitrary-file-read surface on the CLI socket).
- Clamp `maxWidth`/`maxHeight` server-side (<= 1000 px each; 1000x1000 RGBA ≈ 4 MB per response).
- Convert premultiplied → straight alpha; animated images return the first frame.
- Bump `CLI_API_VERSION` 1 → 2 on both sides (`CliApi.kt` and `CliClient.kt` mirror). The CLI treats 404 as "unsupported" and falls back to printing paths, so a new CLI against an old app degrades gracefully.
- Add `CliClient.getRawImage(id, index, maxWidth, maxHeight)` returning pixels + dimensions.

Tests (fast tier):

- Route tests following the `CliServerTest` pattern: dimension clamping, aspect-fit correctness, header/payload size consistency (`w * h * 4`), 404 for missing image/index, first-frame behavior for animated input.

Independent of P1/P3.

guiyanakuang:
Closes #4846. Phase 2 of #4844 (sixel graphics support for Windows Terminal). Independent of the P1 encoder (#4849); the two meet in the P4 wiring (#4848).

## Summary

Adds `GET /cli/paste/{id}/image?index=&maxWidth=&maxHeight=` so the CLI can obtain decoded pixels for sixel rendering without ever decoding image files itself. The app decodes and aspect-fit scales the stored image with its existing Skia stack and returns straight-alpha RGBA8888 with the exact dimensions in `X-Image-Width` / `X-Image-Height` headers. Everything is desktopMain-only — no commonMain changes, no mobile impact.

## Implementation

- **`CliImageTranscoder`** (new, desktopMain): decodes any Skia-supported format (animated images yield their first frame), aspect-fit scales into the requested box without ever upscaling, and reads pixels back as UNPREMUL RGBA_8888 — rendering happens on a premul surface (Skia cannot draw onto unpremul) and the conversion to straight alpha happens in `readPixels`. Undecodable bytes return null (expected input, not an error). Decompression-bomb guard: dimensions are known from the encoded header before pixel decoding, and anything over 64M source pixels is refused.
- **Route** in `CliRouting.kt`: resolves the stored file by paste id + index through `PasteFiles.getFilePaths` (no path parameter — no arbitrary-file-read surface on the CLI socket), caps the input file at 64 MiB, clamps the box to 1000 px per axis (also the default), and answers 404 with a message for every not-servable case: unknown id, no file at the index, unreadable/oversized file, undecodable bytes. The CLI falls back to printing paths on any non-200.
- **`CLI_API_VERSION` 1 → 2** on both sides. A new CLI against an old app degrades gracefully: the route 404s without a server message, which `CliClientException.hasServerMessage` already distinguishes.
- **`CliClient.getRawImage(id, index, maxWidth, maxHeight)`** (CLI side): clamps the requested box, validates the response headers (dimensions in 1..1000) and that the payload is exactly `width * height * 4` bytes — a truncated body is an error, never a partial image.

## Tests

- `CliImageTranscoderTest` (new): original-size passthrough, aspect-fit downscale (100x50 → 10x5), never-upscale, straight-alpha round-trip (half-transparent red must read back r≈0xFF/a≈0x80, catching accidental premul output), undecodable rejection, degenerate box clamp.
- `CliRoutingTest`: 200 with dimension headers and pixel-level payload check against a real PNG on disk, downscale via query params, 404 for unknown paste / out-of-range index / text paste / undecodable stored file.

`./gradlew app:desktopTest` (full fast tier), `cli:macosArm64Test`, and `ktlintCheck` pass.

### DIFF
--- a/app/src/desktopMain/kotlin/com/crosspaste/cli/CliApi.kt
+++ b/app/src/desktopMain/kotlin/com/crosspaste/cli/CliApi.kt
@@ -4,10 +4,14 @@ import kotlinx.serialization.Serializable
 
 /**
  * Version of the local /cli HTTP API served over the Unix domain socket.
- * Bump when the wire contract changes in a way the CLI must know about
- * (no bump needed while the CLI has never shipped in a release).
+ * Bump when the wire contract changes in a way the CLI must know about.
+ * 2: GET /cli/paste/{id}/image raw-RGBA transcode endpoint (sixel support).
  */
-const val CLI_API_VERSION = 1
+const val CLI_API_VERSION = 2
+
+/** Response headers carrying the exact dimensions of a transcoded image. */
+const val CLI_IMAGE_WIDTH_HEADER = "X-Image-Width"
+const val CLI_IMAGE_HEIGHT_HEADER = "X-Image-Height"
 
 const val CLI_ENDPOINT_FILE_NAME = "cli-endpoint.json"
 
--- a/app/src/desktopMain/kotlin/com/crosspaste/cli/CliImageTranscoder.kt
+++ b/app/src/desktopMain/kotlin/com/crosspaste/cli/CliImageTranscoder.kt
@@ -1,0 +1,103 @@
+package com.crosspaste.cli
+
+import org.jetbrains.skia.Bitmap
+import org.jetbrains.skia.ColorAlphaType
+import org.jetbrains.skia.ColorInfo
+import org.jetbrains.skia.ColorType
+import org.jetbrains.skia.Image
+import org.jetbrains.skia.ImageInfo
+import org.jetbrains.skia.Rect
+import org.jetbrains.skia.SamplingMode
+import org.jetbrains.skia.Surface
+
+/**
+ * Decoded pixels for the CLI's sixel rendering (design:
+ * ai/design/cli-sixel-support.md, issue #4846). The CLI never decodes image
+ * files itself; this transcoder is where stored images become raw pixels.
+ */
+class CliRawImage(
+    val width: Int,
+    val height: Int,
+    /** Straight-alpha RGBA8888, row-major, exactly width * height * 4 bytes. */
+    val rgba: ByteArray,
+)
+
+object CliImageTranscoder {
+
+    /** Server-side ceiling for the requested display box, per axis. */
+    const val MAX_BOX_PX = 1000
+
+    /**
+     * Refuses decompression bombs: dimensions are known from the encoded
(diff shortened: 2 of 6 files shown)


guiyanakuang:
Addressed the review findings in 47ced12e:

1. **TOCTOU on the file-size check** — replaced `length()` check + unbounded `readBytes()` with `readImageBytesBounded`: a streaming read capped at limit + 1 bytes, so the 64 MiB budget holds even when a reference-backed file grows or is swapped between check and read. The surrounding `runCatching` is gone with it — only `IOException` is caught now, so JVM resource errors (OOM etc.) propagate instead of masquerading as a 404. Added a route test that grows the stored file past 64 MiB *after* creation and asserts the read-time rejection.

2. **No process-level decode concurrency limit** — a global `Semaphore(1)` (`cliImageTranscodeSemaphore`) now serializes the bounded read + Skia transcode process-wide, so concurrent CLI requests can no longer multiply the ~256 MiB decode peak against the app's heap. Previews are bursty, not throughput-bound; if real concurrency demand ever appears the follow-up is an explicit total-pixel/memory budget, per the review's suggestion.

3. **Missing acceptance tests** — added `animatedGifDecodesToItsFirstFrame` (hand-built 1x1 two-frame GIF89a, first frame red / second blue, asserts red comes back) and server-side clamp coverage at both layers: `clampsOversizedBoxToServerCeiling` (transcoder) and a route test with `maxWidth=999999&maxHeight=999999` against a 2000x100 image asserting the clamped 1000x50 response.

Full fast tier (`app:desktopTest`) and `ktlintCheck` pass.

PR #4851, opened 2026-08-21
:sparkles: add sixel terminal capability probe with DA1 confirmation

### LINKED ISSUE #4847: CLI: terminal capability probe for sixel (DA1 detection + cell pixel size)
Phase 3 of #4844.

Extend terminal image protocol detection with a sixel tier that is confirmed by probing, never assumed from environment variables alone (an older Windows Terminal < 1.22 with `WT_SESSION` set would otherwise receive escape garbage, violating the existing "unrecognized terminals get the path fallback" rule).

Scope:

- `TerminalImage.kt`: add `SIXEL` to `TerminalImageProtocol`. Detection nominates sixel *candidates* from the environment: non-empty `WT_SESSION` (Windows Terminal; conhost does not set it), or `TERM` containing `foot` / `mlterm` / `yaft`. Priority stays KITTY > ITERM > sixel candidates; the tmux/screen guard is unchanged. mintty/WezTerm keep the iTerm protocol (full color beats quantized sixel).
- Probe (only for candidates, and only when both stdin and stdout are TTYs — otherwise fall back to paths): enter raw mode briefly, send `CSI c` (DA1), read the reply with a ~250 ms timeout; attribute `4` in the reply confirms sixel. In the same raw-mode window send `CSI 16 t` (XTWINOPS) to learn the cell pixel size; if unanswered assume 10x20 px.
- New expect/actual platform primitive `platform/TerminalQuery.kt` following the existing 7-file split (`posixNativeMain`: termios tcgetattr/tcsetattr + poll; `mingwNativeMain`: SetConsoleMode with `ENABLE_VIRTUAL_TERMINAL_INPUT`). Do NOT reuse Mordant's `RawModeScope` — it parses input into key events and would consume/mangle the DA response. `pick` must run the probe once *before* entering Mordant raw mode and cache the result.
- Response parsing is pure Kotlin, injectable for tests.

Tests (fast tier):

- Detection matrix: WT_SESSION / foot / mlterm / conhost (no vars) / tmux guard / KITTY-over-sixel priority.
- DA1 reply parser: sixel bit present/absent, malformed replies, timeout → null.
- XTWINOPS reply parser incl. fallback to 10x20.

Independent of P1/P2. Real-terminal behavior is verified in P4.

guiyanakuang:
Closes #4847. Phase 3 of #4844 (sixel graphics support for Windows Terminal). Independent of the P1 encoder (#4849) and the P2 endpoint (#4850); everything meets in the P4 wiring (#4848).

## Summary

Adds the detection layer that decides when sixel may be used, built around one rule: **environment variables only nominate candidates — the terminal itself must confirm sixel via DA1 before a single escape byte of image data is emitted.** This is what keeps an older Windows Terminal (< 1.22, sets `WT_SESSION` but has no sixel) on the existing path fallback instead of receiving escape garbage, preserving the CLI's "unrecognized terminals never get garbage" rule.

## Implementation

- **`TerminalImageProtocol.SIXEL`** added to the enum. The existing env-only detection is unchanged for kitty/iTerm; the two exhaustive `when` renderers get unreachable SIXEL branches guarded at entry (rendering arrives in #4848).
- **`isSixelCandidate(env)`**: `WT_SESSION` non-empty (conhost never sets it), or `TERM` containing `foot` / `mlterm` / `yaft`; the tmux/screen guard outranks every candidate signal.
- **`TerminalProbe.kt`**: `resolveTerminalImageDisplay(stdinInteractive, stdoutInteractive, env, query)` — kitty/iTerm resolve without probing (zero cost); sixel candidates are probed only when both stdin and stdout are interactive. One write carries two queries: XTWINOPS `CSI 16 t` (cell pixel size, may be ignored) then DA1 `CSI c` (universally answered, so its reply doubles as the end-of-transaction marker). Attribute `4` in the DA1 reply confirms sixel; the cell size defaults to 10x20 px when unanswered. All parsers are pure functions; the DA1 parser skips the first parameter (device class, not an attribute) and matches `4` exactly.
- **`queryTerminalRaw`** (per-platform like `SecretInput`, not expect/actual — cliNativeMain resolves it in each target compilation):
  - POSIX: termios raw mode (no ECHO/ICANON, VMIN=0), `poll()`-driven reads against a monotonic deadline, termios restored in `finally`.
  - Windows: `ENABLE_VIRTUAL_TERMINAL_INPUT` so the reply arrives as an ESC character stream, VT output processing so conpty forwards the queries; the read loop drains `ReadConsoleInputW` records (consuming non-key events so they cannot leave the handle signaled forever) and collects key-down characters. Console modes restored in `finally`; mode flags are spelled out locally so compilation does not depend on mingw-w64 header vintage. Deliberately not Mordant's `RawModeScope`, which parses input into key events and would consume the DA reply.

## Tests

18 assertions-worth of fast-tier coverage in `TerminalProbeTest`: the candidate matrix (WT/foot/mlterm/yaft/conhost/tmux guard), DA1 completion detection, DA1 attribute parsing (WT with and without sixel, mlterm, VT100, lone class parameter, exact-match only), cell-size parsing (order, bounds, malformed), and resolver behavior with a recording fake query — kitty resolves without probing, WT+sixel confirms with parsed cell size, old-WT reply falls back, timeout/raw-mode failure falls back, non-interactive streams and non-candidate terminals are never probed.

`cli:macosArm64Test` passes; `cli:compileKotlinLinuxX64`, `cli:compileKotlinLinuxArm64`, and `cli:compileKotlinMingwX64` all compile (validates the per-platform termios/Console API usage); `ktlintCheck` passes. Real-terminal behavior on Windows Terminal is verified in P4 (#4848).

### DIFF
--- a/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/commands/PasteImageOutput.kt
+++ b/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/commands/PasteImageOutput.kt
@@ -88,7 +88,9 @@ internal class InlineImageRenderer(
     }
 
     fun render(paths: List<String>) {
-        val protocol = protocol ?: return
+        // SIXEL renders from app-transcoded pixels, not stored files; it is
+        // wired into this renderer in #4848 and never selected until then
+        val protocol = protocol?.takeIf { it != TerminalImageProtocol.SIXEL } ?: return
         var rendered = 0
         var inspected = 0
         var remainingBudget = maxTotalBytes
@@ -113,6 +115,7 @@ internal class InlineImageRenderer(
             when (protocol) {
                 TerminalImageProtocol.ITERM -> writeItermInlineImage(path.toPath().name, bytes, emit)
                 TerminalImageProtocol.KITTY -> writeKittyInlineImage(bytes, emit)
+                TerminalImageProtocol.SIXEL -> {} // unreachable: filtered out above
             }
             emit("\n")
             rendered++
--- a/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/commands/pick/PickTui.kt
+++ b/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/commands/pick/PickTui.kt
@@ -551,7 +551,10 @@ internal class PickTui(
         path: String,
         panelRow: Int,
     ): Boolean {
-        val protocol = imageProtocol ?: return false
+        // SIXEL renders from app-transcoded pixels, not stored files; it is
+        // wired into this preview in #4848 and never selected until then
+        val protocol =
+            imageProtocol?.takeIf { it != TerminalImageProtocol.SIXEL } ?: return false
         val bytes =
             readImageWithinBudget(
                 path = path,
@@ -581,6 +584,7 @@ internal class PickTui(
                     widthCells = box.first,
                     heightCells = box.second,
                 )
+            TerminalImageProtocol.SIXEL -> {} // unreachable: filtered out above
         }
         terminal.rawPrint("${ESC}8")
         return true
--- a/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/platform/TerminalImage.kt
+++ b/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/platform/TerminalImage.kt
@@ -17,6 +17,13 @@ enum class TerminalImageProtocol {
 
     /** Kitty graphics protocol; PNG payloads only (`f=100`). */
     KITTY,
+
+    /**
+     * Sixel (DCS q): unlike the two protocols above, the payload is quantized
+     * pixels, not a passed-through image file. Never selected from the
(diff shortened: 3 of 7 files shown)


guiyanakuang:
Both points addressed in the follow-up commit:

1. **`wRepeatCount` runs** — `appendKeyChars` now expands each key-down record `repeat(maxOf(wRepeatCount.toInt(), 1))` times, matching Microsoft Terminal's own `ReadConsoleInputStream` behavior, so a compressed run ("22" as one record with `wRepeatCount=2`) can no longer corrupt the DA1 or cell-size reply. A zero count is defensively treated as one so a character is never dropped.

2. **CodeFactor complexity** — `queryTerminalRaw` is split into three helpers: `writeConsolePayload` (payload write + completeness check), `readConsoleReply` (wait/drain loop against the deadline), and `appendKeyChars` (record filtering + repeat expansion). The entry function now only does handle/mode setup and restore.

`cli:compileKotlinMingwX64` and `cli:macosArm64Test` pass (forced rerun).

PR #4852, opened 2026-08-21
:sparkles: wire sixel rendering into paste inline previews and the pick panel

guiyanakuang:
Part of #4844; implements the wiring phase of #4848 (real-hardware verification on Windows Terminal remains to close the issue). Design: `ai/design/cli-sixel-support.md` §2–§4; builds on the merged P1 encoder (#4849), P2 transcode endpoint (#4850), and P3 DA1 probe (#4851).

### `paste` inline preview (`PasteImageOutput.kt`, `PasteCommand.kt`)
- `InlineImageRenderer` now supports SIXEL: instead of passing the stored file through (iTerm/kitty), it renders pixels from an injected `sixelFetch` — the app's `GET /cli/paste/{id}/image` transcode endpoint — keeping the "CLI never decodes image files" rule.
- `PasteCommand` resolves the protocol via `resolveTerminalImageDisplay` (env candidates + DA1 probe confirmation, only when both stdin/stdout are interactive) and requests pixels at terminal width × the probed cell pixel width; height is capped only by the endpoint's 1000 px box clamp, so tall images scroll like text.
- Every fetch failure — including 404 from an app that predates the endpoint — falls back to the file paths already printed above, counted in the existing "not previewed" note. The 20 MiB total budget now counts fetched RGBA bytes for sixel.

### `pick` preview panel (`PickCommand.kt`, `PickTui.kt`)
- The sixel probe runs once per command, BEFORE Mordant's raw mode (its key-event parser would eat the DA1 reply), and the cached `TerminalImageDisplay` is reused across Ctrl-E editor round trips.
- The sixel fetch is async on `fetchScope` like the list/preview fetches, so a slow transcode cannot freeze the key loop; the reply draws via a new `FetchMsg.SixelPixels` message. Stale replies (selection moved, panel resized, repaint pending) are dropped — the superseding repaint has already rescheduled a fresh draw.
- The app scales pixels to exactly the panel cell box × probed cell pixel size, so the reply never occupies more rows than the panel reserves. Like iTerm, sixel pixels live in cells: repaints erase them, so the retransmit condition became `protocol != KITTY`.

### Hardening (from review)
- `CliClient.getRawImage` now wraps response-body reads: a mid-transfer socket failure surfaces as `CliClientException` per its documented contract instead of a raw ktor exception tearing down the command/picker.

## Tests
- `PasteImageOutputTest` adapted to the now-suspend `render` (`runTest`) and extended with sixel cases: per-index fetch + emitted DCS framing, fetch-failure fallback note, and RGBA-byte budget accounting.
- Existing iTerm/kitty ceilings and detection/probe suites unchanged and green.
- Verified: `cli:macosArm64Test`, `mingwX64`/`linuxX64` cross-compiles, `ktlintCheck`.

## Remaining for #4848
Real-hardware verification on Windows Terminal ≥ 1.22 (Parallels VM): `paste` single/multi-image, transparent PNG, JPEG; `pick` preview draw/erase/scroll behavior; degradation matrix (conhost, piped stdout, old app + new CLI).

### DIFF
--- a/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/api/CliClient.kt
+++ b/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/api/CliClient.kt
@@ -173,7 +173,7 @@ class CliClient(
                 "/cli/paste/$id/image?index=$index&maxWidth=$boxWidth&maxHeight=$boxHeight",
             )
         if (!response.status.isSuccess()) {
-            val serverMessage = extractMessage(response.bodyAsText())
+            val serverMessage = readBodyForImage { extractMessage(response.bodyAsText()) }
             throw CliClientException(
                 serverMessage ?: "Request failed with status ${response.status}",
                 statusCode = response.status.value,
@@ -189,7 +189,9 @@ class CliClient(
         ) {
             throw CliClientException("Unexpected image dimensions from CrossPaste: $width x $height")
         }
-        val rgba = response.readRawBytes()
+        val rgba =
+            readBodyForImage { response.readRawBytes() }
+                ?: throw CliClientException("Failed to read the image payload from CrossPaste.")
         if (rgba.size != width * height * 4) {
             throw CliClientException(
                 "Image payload size ${rgba.size} does not match $width x $height RGBA.",
@@ -197,6 +199,22 @@ class CliClient(
         }
         return CliRawImage(width, height, rgba)
     }
+
+    /**
+     * Body reads happen after [request] returns, so a mid-transfer socket
+     * failure would otherwise throw a raw ktor exception — escaping the
+     * documented contract that every [getRawImage] failure is a
+     * [CliClientException] (callers catch it and fall back to printing
+     * paths). Null means the read failed.
+     */
+    private suspend fun <T> readBodyForImage(read: suspend () -> T): T? =
+        try {
+            read()
+        } catch (e: CancellationException) {
+            throw e
+        } catch (_: Exception) {
+            null
+        }
 
     /**
      * Opens a long-lived streaming GET (the /cli/watch NDJSON feed) and hands
--- a/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/commands/PasteCommand.kt
+++ b/cli/src/cliNativeMain/kotlin/com/crosspaste/cli/commands/PasteCommand.kt
@@ -1,7 +1,11 @@
 package com.crosspaste.cli.commands
 
 import com.crosspaste.cli.CliContext
-import com.crosspaste.cli.platform.detectTerminalImageProtocol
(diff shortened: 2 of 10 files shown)


guiyanakuang:
Follow-up 2e2da64c addressing review feedback:

- **Pixel-box acceptance tests (was P2)**: the cell-box → pixel-box computation is now a pure `sixelPixelBox(columns, rows, display)` in TerminalProbe.kt, using Long intermediates and clamping both axes to `CLI_IMAGE_MAX_BOX_PX` with a floor of 1. Both call sites use it (pick's panel box; paste passes `rows = CLI_IMAGE_MAX_BOX_PX` so the height saturates at the endpoint clamp — paste deliberately has no row limit). New tests cover the default 10×20 cell, probed metrics, per-axis clamping, Int-overflow saturation, and degenerate (zero/negative) cell counts.
- **Explicit fetch cancellation (P1 addendum)**: the blocking-fetch issue was already fixed in the initial commit (fetch runs on `fetchScope`, replies return through the results channel with id + cell box, stale replies are dropped; `tick()` is not suspend). This commit adds the remaining suggestion: a superseded or no-longer-needed fetch is now actively cancelled (`sixelJob`) instead of running to completion — on relaunch for a new paste and when the candidate disappears (selection moved off an image / preview closed). Cancelling always clears the in-flight marker so re-selecting the same paste can re-fetch.

guiyanakuang:
Follow-up 88b6c188: the in-flight guard for pick's sixel fetch was keyed by paste id only, which loses the post-resize redraw — resize during an in-flight fetch reschedules a draw for the new box, the due draw is swallowed by the id-only guard (consuming the deadline), and the old-box reply is then discarded as geometry-stale: no request, no deadline, image gone until an unrelated repaint. Same root cause also let a cancelled request's reply race back as if current, and let a mid-flight repaint's leftover deadline relaunch a duplicate transcode after a successful draw.

Fix per review: requests are now identified by a full `SixelRequestKey(generation, pasteId, boxColumns, boxRows)`, managed by a pure `SixelFetchState` (new file, unit-testable):

- `nextRequest` reuses the in-flight request only on an EXACT match (paste and box); a same-paste/new-box request supersedes it (caller cancels the old job).
- `FetchMsg.SixelPixels` carries the full key; only the reply equal to the active key may clear the active state or draw — superseded boxes/pastes and cancelled generations are dropped without touching it.
- A geometry-stale-but-active reply (resize whose refetch deadline has not fired yet) skips drawing; the pending deadline redraws.
- A successful draw clears `imageDeadline`, so a repaint that occurred mid-flight no longer schedules a redundant refetch+transcode.

`SixelFetchStateTest` covers the three requested scenarios: resize-while-in-flight (new box launches, old reply stale), cancel-then-reselect same paste (generation keeps the old reply out), and identical-in-flight reuse with exactly-once completion, plus selection-change supersession.

PR #4861, opened 2026-08-21
:hammer: replace BouncyCastle SPAKE2 EC backend with constant-time OpenSSL via JNA

### LINKED ISSUE #4860: Replace BouncyCastle SPAKE2 EC backend with a constant-time OpenSSL backend and remove bc-java
## Background

Pairing v3 uses SPAKE2 over P-256 (RFC 9382). The 2026-07-24 constant-time review (#4662) found that BouncyCastle's P-256 point addition/doubling formulas contain input-dependent branches, so it cannot satisfy RFC 9382 §7; production has been fail-closed (`UnavailablePakeProvider`) since then, and BouncyCastle was retained for test vectors only.

RFC 9382 §7 is the gate for ever flipping production `PAIRING_VERSION` to 3, so the desktop JVM needs a reviewed constant-time EC backend. OpenSSL's `EC_POINT_mul` is documented to run in constant time for scalars in `[0, n)` and ships the cache-safe `ecp_nistz256` P-256 implementation.

## Proposal

1. Add `OpenSslPakeEcOps` (core desktopMain): a JNA binding to OpenSSL 3 libcrypto's low-level EC/BN API implementing `PakeEcOps`. Scalars are validated into `[1, n)` before every multiplication (the constant-time precondition), secrets are released with `BN_clear_free`, and library candidates are probed for the full required symbol set before being accepted (rejects e.g. Apple's system LibreSSL). Resolution: `crosspaste.libcrypto.path` / `CROSSPASTE_LIBCRYPTO_PATH` override first, then per-platform well-known names.
2. Remove bc-java entirely (backend class, Gradle dependency, version catalog entry) and point every consumer — RFC 9382 vector tests, provider contract tests, pairing v3 integration tests, the e2e `HeadlessPeer`, and the DEVELOPMENT interop DI path — at the OpenSSL backend, so tests and production share one backend.

## Out of scope

- Production stays fail-closed: `PAIRING_VERSION` remains 2 and production DI still registers `UnavailablePakeProvider`. No user-visible behavior change.
- Bundling libcrypto into the distributable (Conveyor) is deferred to the v3 rollout work; dev/test environments resolve the library from the environment (Homebrew on macOS, distro libssl3 on Linux CI).

guiyanakuang:
Closes #4860

## Summary

- Add `OpenSslPakeEcOps` (core desktopMain): JNA binding to OpenSSL 3 libcrypto's low-level P-256 ops (`EC_POINT_mul/add/invert/oct2point/point2oct`, `BN_nnmod`, `BN_priv_rand_range`, …) implementing the `PakeEcOps` boundary.
  - Every scalar is validated into `[1, n)` before any multiplication — the documented precondition for OpenSSL's constant-time path — and only single-point `EC_POINT_mul` is used.
  - Secret scalars are released with `BN_clear_free` (zeroized), improving on the immutable `BigInteger` copies the BouncyCastle backend left to the GC.
  - Library resolution: `crosspaste.libcrypto.path` system property / `CROSSPASTE_LIBCRYPTO_PATH` env override first, then per-platform well-known names. Each candidate is probed for the full required symbol set before acceptance, so libraries missing the OpenSSL 1.1.1+ surface (e.g. Apple's system LibreSSL) are rejected with a clear `PakeException` instead of an `UnsatisfiedLinkError` mid-operation. The loaded instance is a memoized process-wide singleton; a failed load is retried on next access.
- Remove bc-java entirely: delete `BouncyCastlePakeEcOps`, drop the `bcprov-jdk18on` dependency and catalog entries. All consumers now use the OpenSSL backend — RFC 9382 vector tests, `Spake2PakeProviderTest`, `PairingV3IntegrationTest`, e2e `HeadlessPeer`, and the DEVELOPMENT interop DI path (which falls back fail-closed to `UnavailablePakeProvider` with a warning when libcrypto is unavailable). Tests and production now share one backend.
- Update stale backend references in KDoc (`PakeEcOps`, `PakeProvider`, `NoblePakeEcOps`, vector test base class).

## Production behavior

Unchanged. `PAIRING_VERSION` remains 2 and production DI still registers `UnavailablePakeProvider`; no production code path loads libcrypto. Bundling libcrypto into installers is deferred to the v3 rollout work.

## Test plan

- `Spake2Rfc9382VectorTest` (RFC 9382 Appendix B official vectors, byte-for-byte) now runs against the OpenSSL backend: 3/3 green.
- `Spake2PakeProviderTest` contract suite (12), `PairingV3IntegrationTest` (25, `-PintegrationTests`), `DesktopPairingCapabilityTest` (5): green.
- Full fast tier `core:desktopTest` + `app:desktopTest` and `:e2e:compileKotlinDesktop`: green. `ktlintCheck` clean.
- CI runners provide libcrypto out of the box (ubuntu: distro `libssl3`; macos-26: Homebrew openssl@3).


guiyanakuang:
Follow-up commit b9187cb0 addresses two review findings:

**1. Constant-time guarantees were incomplete (P1).** `BN_nnmod` on the PIN-derived 48-byte HKDF output ran without `BN_FLG_CONSTTIME`, and the scalar range check used variable-time `BN_cmp`/`BN_is_zero` on secret scalars. Fixed: the binding now includes `BN_set_flags` and sets `BN_FLG_CONSTTIME` on the wide input before reduction so division takes OpenSSL's no-branch path, and the `[1, n)` range validation is now a branchless byte-level comparison against the cached order bytes (single public valid/invalid branch only; `BN_cmp` removed from the binding). The class KDoc no longer claims "reviewed" — final sign-off explicitly rests with the independent security review.

**2. Capability could outrun the backend (P1).** The advertised pairing capability was resolved before and independently of the PAKE provider, so a failed libcrypto load left the device advertising v3 while the provider was `UnavailablePakeProvider` — peers would select V3_PIN and fail at the PAKE instead of negotiating v2. Fixed: `resolveDesktopPairingBackend` now resolves "requested + backend availability" once into an immutable `DesktopPairingBackend` (capability + provider), registered as a single Koin snapshot that both consumers derive from; a load failure falls back to advertising v2. New test `developmentBackendLoadFailureFallsBackToAdvertisingV2` covers the failure path via an injectable loader, and the remaining capability tests now assert the loader is never invoked on non-v3 paths. The old `providerStillFailsClosedForMisinjectedProductionCapability` test was removed because the misinjection it guarded against is now unrepresentable by construction.

Full fast tier, `PairingV3IntegrationTest` (25, `-PintegrationTests`), and RFC 9382 vectors: green.

PR #4863, opened 2026-08-21
:construction_worker: bundle pinned OpenSSL libcrypto into desktop packages via Conveyor

### LINKED ISSUE #4862: Bundle OpenSSL libcrypto with desktop packages for the pairing v3 SPAKE2 backend
## Background

The pairing v3 SPAKE2 backend (`OpenSslPakeEcOps`, #4860) loads OpenSSL 3 libcrypto via JNA. Today no CrossPaste distribution bundles libcrypto: dev/test/CI resolve it from the environment (Homebrew, distro `libssl3`), and production keeps v3 fail-closed (`UnavailablePakeProvider`), so nothing breaks — but v3 cannot ship until the app carries its own reviewed libcrypto on every desktop platform. macOS and Windows have no system OpenSSL at all, and depending on unvetted distro libraries is unacceptable for a cryptographic component.

Binaries are built by our own CI at [CrossPaste/openssl-builds](https://github.com/CrossPaste/openssl-builds) (fork of whyoleg/openssl-builds; build scripts unchanged, sources from the official OpenSSL tarball). Release [3.6.0](https://github.com/CrossPaste/openssl-builds/releases/tag/3.6.0) covers all five desktop targets with published SHA256 checksums; the `ecp_nistz256` constant-time P-256 assembly and required symbol exports were verified per target, and the RFC 9382 vector suite passes against the built macos-arm64 dylib.

## Scope (packaging only)

1. Pin per-target download URL + SHA-256 in `app/openssl.yaml` (same pattern as `jbr.yaml`).
2. A `prepareOpenSslLibs` Gradle task: download each pinned library, verify SHA-256 (self-healing on corrupted cache), and wrap it into a thin jar whose only entry is the library under its runtime file name. Optional `-PopenSslTargets=` subset for machine-scoped CI builds.
3. Add the five jars as per-machine Conveyor inputs so `extract-native-libraries` places the library in the JVM lib directory (= `skiko.library.path`) and signs/notarizes it — the same proven pipeline as the skiko/tesseract platform jars.
4. Wire the task + a cache keyed on `openssl.yaml` into the release and beta workflows.

Out of scope (follow-ups): tightening the runtime resolution order in `OpenSslPakeEcOps` (bundled path first, no system fallback in production), real-machine load verification across all five packages, and the eventual `PAIRING_VERSION` bump. Production behavior is unchanged by this issue.

guiyanakuang:
Closes #4862

## What

Bundles our own CI-built OpenSSL 3.6.0 libcrypto with every desktop package, as the packaging prerequisite for enabling pairing v3. Production behavior is unchanged: the runtime still resolves libcrypto exactly as before, and v3 stays fail-closed behind `PAIRING_VERSION=2`.

- **`app/openssl.yaml`** — pins per-target download URL + SHA-256 for the five desktop libraries, sourced from the [CrossPaste/openssl-builds 3.6.0 release](https://github.com/CrossPaste/openssl-builds/releases/tag/3.6.0) (fork of whyoleg/openssl-builds with unchanged build scripts; per-target `ecp_nistz256` assembly and symbol exports verified, RFC 9382 vectors green against the built macos-arm64 dylib).
- **`prepareOpenSslLibs` Gradle task** — downloads each pinned library, verifies SHA-256 (re-downloads on corrupted cache, hard-fails on mismatch), prunes stale files from earlier versions, and wraps each library into a thin jar whose only entry is the library under its runtime file name (`libcrypto.3.dylib` / `libcrypto.so.3` / `libcrypto-3-x64.dll`). `-PopenSslTargets=` selects a subset so machine-scoped CI builds don't depend on other platforms' binaries.
- **`conveyor.conf`** — adds the five jars as per-machine inputs. Conveyor's `extract-native-libraries` step places the library into the JVM lib directory (= `skiko.library.path`) and signs/notarizes it, the same pipeline already proven by the skiko/javacpp/tesseract platform jars.
- **CI** — release and beta workflows run the task before the Conveyor build with a cache keyed on `openssl.yaml` (the linux-only beta uses a distinct key and the linux target subset).

## Verification

- Full run downloads all five libraries and produces the five jars; jar entry round-trips to the pinned SHA-256.
- Second run is `UP-TO-DATE`; a tampered cached library is detected and re-downloaded; a stale old-version file is pruned; an unknown `-PopenSslTargets` name fails fast.
- `conveyor json` on the local mac config resolves `crosspaste-libcrypto-macos-arm64.jar` into `mac.aarch64` inputs (HOCON valid, path correct).
- `ktlintCheck` and `writeConveyorConfig -PappEnv=PRODUCTION` pass.

## Out of scope (follow-ups)

Runtime resolution tightening in `OpenSslPakeEcOps` (bundled path first, no system-path fallback in production), real-machine load verification of all five packages, and the `PAIRING_VERSION` bump.

### DIFF
--- a/.github/workflows/build-beta-linux.yml
+++ b/.github/workflows/build-beta-linux.yml
@@ -95,6 +95,14 @@ jobs:
           path: ./app/jbr
           key: jbr-${{ hashFiles('**/jbr.yaml') }}
 
+      - name: Cache OpenSSL libcrypto binaries
+        uses: actions/cache@v5
+        with:
+          path: ./app/openssl
+          # Linux-only subset (see prepareOpenSslLibs below), so use a key
+          # distinct from the release workflow's full five-target cache.
+          key: openssl-linux-${{ hashFiles('**/openssl.yaml') }}
+
       - name: Build with Gradle
         run: ./gradlew app:build -PintegrationTests
 
@@ -108,6 +116,9 @@ jobs:
         run: |
           VERSION=$(grep '^version=' app/src/desktopMain/resources/crosspaste-version.properties | cut -d= -f2)
           ./gradlew :cli:assembleDistLinuxX64 :cli:assembleDistLinuxArm64 "-PcliVersion=${VERSION}.${REVISION}"
+
+      - name: Prepare OpenSSL libcrypto jars
+        run: ./gradlew app:prepareOpenSslLibs "-PopenSslTargets=linux-x64,linux-arm64"
 
       - name: Prepare Conveyor config (Linux only)
         run: |
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -197,6 +197,12 @@ jobs:
           path: ./app/jbr
           key: jbr-${{ hashFiles('**/jbr.yaml') }}
 
+      - name: Cache OpenSSL libcrypto binaries
+        uses: actions/cache@v5
+        with:
+          path: ./app/openssl
+          key: openssl-${{ hashFiles('**/openssl.yaml') }}
+
       - name: Restore dylib cache for x86_64
         uses: actions/cache/restore@v5
         with:
@@ -255,6 +261,9 @@ jobs:
           else
             echo "Directory does not exist!"
           fi
+
+      - name: Prepare OpenSSL libcrypto jars
+        run: ./gradlew app:prepareOpenSslLibs
 
       - name: Prepare Dependency
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
       app: ${{ steps.filter.outputs.app }}
       cli: ${{ steps.filter.outputs.cli }}
       web: ${{ steps.filter.outputs.web }}
+      openssl: ${{ steps.filter.outputs.openssl }}
     steps:
       - uses: actions/checkout@v6
       - uses: dorny/paths-filter@v4
@@ -59,6 +60,13 @@ jobs:
               - 'settings.gradle.kts'
               - 'build.gradle.kts'
               - 'gradle.properties'
+              - '.github/workflows/ci.yml'
+            # Bundled libcrypto pin: verify the download chain on the PR
+            # instead of discovering a bad URL/SHA at release packaging time.
+            openssl:
+              - 'app/openssl.yaml'
+              - 'conveyor.conf'
+              - 'app/build.gradle.kts'
               - '.github/workflows/ci.yml'
 
   app-build:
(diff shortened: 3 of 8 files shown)


guiyanakuang:
Addressed the review feedback in ffba5080. All four points were verified against primary sources before changing anything:

**P1 — known-vulnerable 3.6.0 → pinned 3.6.3.** Confirmed via the [OpenSSL vulnerabilities list](https://openssl-library.org/news/vulnerabilities/): 3.6.0 accumulates ~40 CVEs including two High (CVE-2025-15467, fixed in 3.6.1; CVE-2026-45447, fixed in 3.6.3). One correction to the review: 3.6.4 is not released — the newest tag in openssl/openssl is `openssl-3.6.3`, so that is the pin (the 3.6.4 references on the vulnerabilities page are two Low CVEs fixed in the unreleased next patch). Rebuilt via the fork CI in a single run ([32547398779](https://github.com/CrossPaste/openssl-builds/actions/runs/32547398779), all four platform jobs plus aggregate green after folding the `vs_version=18` fix into the main workflow), published as [openssl-builds 3.6.3](https://github.com/CrossPaste/openssl-builds/releases/tag/3.6.3), and re-verified end to end: embedded `OpenSSL 3.6.3 9 Jun 2026` version string, `ecp_nistz256` assembly on all five targets, exported symbol set (windows export table parsed: 5895 names), macOS `minos 11.0`, and the RFC 9382 vector + provider suite (15 tests) green against the built macos-arm64 dylib. `openssl.yaml` re-pinned with the new per-target SHA-256s.

**P1 — OpenSSL license now ships in every package.** `app/third-party/licenses/openssl/LICENSE.txt` (Apache-2.0, taken from the `openssl-3.6.3` tag) is added to the generic Conveyor inputs, landing at `third-party/licenses/openssl/LICENSE.txt` inside the app directory on all platforms.

**P2 — jar wrapping removed.** Confirmed in the Conveyor JVM docs that bare shared libraries in inputs are moved to the shared library directory, so the task now stages plain files at `app/openssl/<target>/<runtime lib name>` and `conveyor.conf` references them directly. This also removes the ZIP-timestamp reproducibility problem the review demonstrated. Verified locally with `conveyor json`: the bare dylib and the license directory both resolve into `mac.aarch64` inputs.

**P2 — PR CI now validates the pin.** New `openssl-verify` job in `ci.yml`, gated by a paths filter (`app/openssl.yaml`, `conveyor.conf`, `app/build.gradle.kts`, the workflow itself): it runs `prepareOpenSslLibs` (download + SHA-256, hard fail on mismatch) and then asserts every `app/openssl/...` path referenced by `conveyor.conf` exists. Wired into the `ci` summary gate.

Also exercised the version-bump path for real: re-running the task after the 3.6.0 → 3.6.3 pin change pruned the stale 3.6.0 downloads and staged the new libraries with matching hashes.

PR #4869, opened 2026-08-22
:construction_worker: auto-download the pinned libcrypto for development runs so pairing v3 loads without a system OpenSSL

### LINKED ISSUE #4868: Development runs cannot load pairing v3: auto-download the pinned libcrypto before app:run
## Problem

`./gradlew app:run` on a machine without a system OpenSSL 3 (typical on Windows) starts fine but logs `OpenSSL libcrypto unavailable, pairing v3 stays disabled` and falls back to advertising pairing v2. The pinned libcrypto (app/openssl.yaml) is only downloaded by the packaging-time task `prepareOpenSslLibs`, so development runs depend on whatever OpenSSL happens to be installed in the environment — on Windows usually nothing — which blocks local pairing v3 interop testing right when v3 development is being pushed toward release.

## Proposal

Mirror the JBR auto-download: before a development run launches, download and stage the current platform's pinned libcrypto and point the run JVM at it via the existing `crosspaste.libcrypto.path` override (honored by `LibCryptoResolution.Environment` only).

- New `prepareDevOpenSslLib` task stages only the current platform's target (SHA-256 verified, cached once downloaded), instead of coupling dev startup to all five platform binaries.
- `run` / `desktopRun` / hot-reload tasks depend on it and receive the override system property.
- An explicit `CROSSPASTE_LIBCRYPTO_PATH` environment variable still wins (troubleshooting escape hatch).
- Packaged builds and tests are untouched: only run tasks get the property, and `Bundled` (production/beta) resolution ignores overrides by design.

guiyanakuang:
Closes #4868

## Summary

`./gradlew app:run` on a machine without a system OpenSSL 3 (typical on Windows) silently falls back to pairing v2 (`OpenSSL libcrypto unavailable, pairing v3 stays disabled`), blocking local pairing v3 interop testing. The pinned libcrypto (app/openssl.yaml) was previously downloaded only by the packaging-time `prepareOpenSslLibs` task.

This wires the JBR-style automatic download into development runs:

- **`stageOpenSslTarget` extracted**: the download + SHA-256 self-healing verification + bare-library staging logic from `prepareOpenSslLibs` is now a shared function; the packaging task's behavior is unchanged.
- **New `prepareDevOpenSslLib` task**: downloads and stages only the current platform's target (~5MB), so development startup is not coupled to the reachability of all five platform binaries. Once staged with a matching SHA-256, subsequent runs skip the network entirely.
- **Run-task wiring**: `run` / `desktopRun` / hot-reload tasks depend on the new task and receive `-Dcrosspaste.libcrypto.path=<staged path>` — the existing explicit override honored by `LibCryptoResolution.Environment`.

Deliberate boundaries:

- An explicit `CROSSPASTE_LIBCRYPTO_PATH` environment variable suppresses the injected property, preserving the documented troubleshooting escape hatch.
- Only run tasks get the override; packaged builds and tests are unaffected. `Bundled` (production/beta) resolution ignores overrides by design, so no startup warning can appear in shipped builds.

## Verification

- `:app:run --dry-run` shows `:app:prepareDevOpenSslLib` in the task graph.
- Ran `prepareDevOpenSslLib` on macOS arm64: 3.6.3 library downloaded, SHA-256 verified, staged at `app/openssl/macos-arm64/libcrypto.3.dylib`.
- `core:desktopTest` with `CROSSPASTE_LIBCRYPTO_PATH` pointed at the staged library: RFC 9382 vectors 3/3 and Spake2 provider tests 12/12 pass with zero skips — the staged binary loads and is cryptographically correct.
- `prepareOpenSslLibs -PopenSslTargets=macos-arm64` (packaging path) still succeeds after the refactor.
- `app:run` confirmed to be a `JavaExec` task, so `systemProperty` reaches the launched JVM.
- `ktlintFormat` clean.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -2,6 +2,7 @@
 import org.gradle.api.tasks.testing.logging.TestLogEvent
 import org.gradle.nativeplatform.platform.internal.DefaultNativePlatform
 import org.jetbrains.compose.desktop.application.dsl.TargetFormat
+import org.jetbrains.compose.reload.gradle.AbstractComposeHotRun
 import org.jetbrains.compose.reload.gradle.ComposeHotRun
 import org.yaml.snakeyaml.LoaderOptions
 import org.yaml.snakeyaml.Yaml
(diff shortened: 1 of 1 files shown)


guiyanakuang:
Addressed all review findings in 49c68c67:

1. **All hot-reload entry points wired.** The sync variants (`hotRunDesktop`, `hotDevDesktop`, deprecated alias) are now configured via their common base `AbstractComposeHotRun` (a `JavaExec` subclass), and the async variants (`hotRunDesktopAsync`, `hotDevDesktopAsync`) — whose `ComposeHotAsyncRun` type is internal to the plugin — get the staging dependency via a name match. Verified with `--dry-run`: all four hot tasks now include `:app:prepareDevOpenSslLib`, and the generated `desktopMain.argfile` carries the injected `crosspaste.libcrypto.path` property, so async launches get both the staged library and the pointer to it.
2. **Explicit overrides win and skip the download.** Resolution priority is now `-Dcrosspaste.libcrypto.path` (on the Gradle invocation, forwarded to the launched JVM since Gradle JVM properties do not propagate on their own) > `CROSSPASTE_LIBCRYPTO_PATH` environment variable (propagates natively, nothing injected) > automatically staged library. When either override is present, `prepareDevOpenSslLib` is not added to the graph, so an offline machine with its own libcrypto starts normally. Verified via `--dry-run` for both override forms.
3. **Lazy manifest parsing.** `openssl.yaml` is now parsed behind a `lazy` that is only touched when a development run task is actually configured; `:app:desktopTest --dry-run` shows no staging task and no manifest read.
4. **Explicit os/arch mapping.** `aarch64/arm64` and `amd64/x86_64/x64` map explicitly; any other arch, and Windows ARM64, yield no target. A development run on such a platform logs a clear warning and starts without wiring (same behavior as before this PR — environment resolution), rather than failing startup; invoking `prepareDevOpenSslLib` directly fails with an explicit error naming os.name/os.arch. Chose warn-and-skip for the run path deliberately so unsupported platforms can still launch the app.
5. **DEVELOPMENT-only.** The whole wiring is gated on `appEnv` (default DEVELOPMENT); `-PappEnv=PRODUCTION`/`BETA` runs get neither the download nor the property, verified via `--dry-run`.

Full verification matrix (all `--dry-run` on task graphs): `run` / `hotRunDesktopAsync` / `hotDevDesktop` / `hotDevDesktopAsync` → staging task present; `desktopTest`, `-D` override, env override, `-PappEnv=PRODUCTION` → absent. `prepareDevOpenSslLib` and `prepareOpenSslLibs -PopenSslTargets=macos-arm64` both still pass, `ktlintFormat` clean.

One pre-existing gap noticed while fixing (not introduced here, left untouched): `initJvmArgs` and the `compileSwift` dependency are wired via `withType<ComposeHotRun>`, which does not cover `hotDevDesktop`/async — those entry points launch without `-DappEnv=DEVELOPMENT` and the other base JVM args. Worth a separate issue if the hot-dev entry points are actually used.

PR #4959, opened 2026-09-09
:sparkles: mark the local device row with plain text after its name instead of a status tag

### LINKED ISSUE #4958: Local device row: render "current device" as plain text after the name, not a status tag
## Problem

`LocalDeviceRow` (#4957) puts a `StateTagView` ("CURRENT DEVICE" pill with icon, uppercase) into the `nameTrailing` slot of `DeviceRowContent`. That slot was meant for a short inline marker that reads as part of the title; a status-tag chip next to the name looks like a sync state, is wide, and cannot flow like text.

## Fix

Add a small `DeviceNameMarker(text)` composable next to `DeviceRowContent`: plain `labelMedium` text in the primary color, single line, no background or icon. `LocalDeviceRow` uses it with the existing `current_device` key. Mobile's local device card will use the same marker with its own key so both platforms look alike.

guiyanakuang:
Closes #4958

## What

- New `DeviceNameMarker(text)` in `DeviceRowContent.kt`: bold `labelMedium` text in `colorScheme.primary`, single line, no container or icon. It is the intended content for the `nameTrailing` slot.
- `LocalDeviceRow` passes `DeviceNameMarker(copywriter.getText("this_device"))` instead of a `StateTagView`.
- New i18n key `this_device` in all 10 locales, using the native phrase for the device you are on: "This device", "本机", "本機", "このデバイス", "이 기기", "Dieses Gerät", "Cet appareil", "Este dispositivo", "این دستگاه".

## Why

The "CURRENT DEVICE" chip reads like a sync-state tag and is much wider than the words. As plain text it belongs to the title: beside a short name, or on the line below a long one (the name line is a FlowRow since #4955).

An earlier revision used a bare upper-cased "current". That is not the idiom device lists use (Apple, Google, WhatsApp, LocalSend all say "This device"), and its one-word translations read as time adverbs in CJK or as "up to date" in German ("Aktuell"). Upper-casing was also a no-op for CJK and Persian, so the emphasis differed by locale; the marker is now regular case and bold.

Mobile will reuse `DeviceNameMarker` and the `this_device` key for its local device card.

## Verification

`./gradlew ktlintFormat` / `ktlintCheck` clean, `app:compileKotlinDesktop` passes. Not yet verified on a running build.


guiyanakuang:
Restyled per review: the marker is now a single upper-cased word ("CURRENT"), bold with 1sp tracking in the primary color. New i18n key `current` (Current / 当前 / 目前 / 現在 / 현재 / Aktuell / Actuel / Actual / Atual / فعلی); `current_device` stays in use by the dialog.

Issue #4996, opened 2026-09-15
The local area network built by Windows via Internet Connection Sharing cannot be found in network settings.


happyOY:
### How frequently does the bug occur?

Always

### Description

When I set up a new local area network using Windows Internet Connection Sharing,
Windows' LAN address is set to 192.168.137.1. At the same time, another address connected to the home router is 192.168.101.128.
The secondary machine on the other end connects to the subnet created by Windows instead of the home router, with an IP address of 192.168.137.8.
A strange issue occurs at this point: the secondary machine (192.168.137.8) cannot establish a connection to Windows (192.168.137.1).
However, Windows (192.168.137.1) can successfully connect to the secondary machine at 192.168.137.8. 
Naturally, clipboard content can only be transmitted normally from Windows to the secondary machine.

Later I found an issue: when I entered the IP address of the Windows home router (192.168.101.128) on the secondary device, the connection could be established. Strangely, the secondary device was not connected to the home router's network.

We hope the developers can fix this bug, so clipboard interaction can be achieved through a direct network cable connection without a router.

### Stacktrace & log output

```shell

```

### Can you reproduce the bug?

Always

### Reproduction Steps

1. Enable Internet Connection Sharing on Windows to create a sub-LAN
2. Connect the secondary Linux machine to the Windows sub-LAN directly via an Ethernet cable
3. Enter the peer IP address to reproduce this bug

### Version

v2.2.0(2544)

### OS

windows and linux

guiyanakuang:
Tracked down — thanks for the report. The "entering the home router's IP works instead" detail is exactly what pointed at the cause.

**Cause**

CrossPaste auto-binds only **one** network interface at startup, and in your setup it always picks the wrong one: when everything else ties, the sort prefers the higher last octet, and an ICS gateway address is always pinned to `.1`. So it binds `192.168.101.128` — precisely the interface your secondary machine cannot see.

When that machine dials `192.168.137.1`, Windows accepts the connection but replies with an **empty** address list, because that address isn't in the bound set. The peer then stores a device with no address at all and can never connect to it, while the dialog still reports success. That's our bug, not anything you did wrong.

Entering `192.168.101.128` works because your secondary machine's default gateway *is* the Windows box, so it can route into that subnet — and that address is in the bound set, so the list comes back non-empty.

**Fixed**

[...]

PR #5009, opened 2026-09-16
:bug: refresh the network settings interface list on network changes

### LINKED ISSUE #5008: Network settings interface list never refreshes after the page is first opened
### Description

`Settings → Network` enumerates the interface list exactly once, in a `LaunchedEffect(Unit)` when the page composes:

```kotlin
LaunchedEffect(Unit) {
    networkInterfaces = networkInterfaceService.getSortedNetworkInterfaceInfo()
}
```

So an interface that comes up *after* that first composition is never offered. Plug in an Ethernet cable, switch on a hotspot, or enable Windows Internet Connection Sharing while the app is running, and the picker keeps showing the interfaces that existed when the page was first opened — with no way to tick the new one short of restarting the app.

That matters because the picker is the only workaround for a multi-homed host today: auto-select binds exactly one interface, and on an ICS/hotspot host it systematically picks the wrong one (see #4996, #4997).

Two smaller problems in the same spot:

- The enumeration is a blocking `NetworkInterface.getNetworkInterfaces()` call made on the composition dispatcher.
- The list never reflects a runtime IP change (DHCP renewal), even though `NetworkStateMonitor` already reports one to the discovery layer.

### Expected

The picker tracks the live interface set: it re-reads on every OS network change while the page is open, and re-reads when the page is opened again.

### Notes

`NetworkInterfaceService.networkInterfaces` cannot serve this — it carries only the *bound selection*, and it is deliberately `distinctUntilChanged`, so an interface appearing outside the selection produces no emission at all. The picker needs the full live set.

guiyanakuang:
Closes #5008.

`Settings → Network` read the interface list once, in a `LaunchedEffect(Unit)`, so an interface that came up after the page was first composed was never offered — you had to restart the app to tick a hotspot / Windows ICS adapter. That picker is currently the only workaround for a multi-homed host (#4996), so a stale list is the difference between a working workaround and none.

**What changed**

- `NetworkInterfaceService` gains `allNetworkInterfaces: Flow<List<NetworkInterfaceInfo>>` — the full live interface set, as opposed to `networkInterfaces`, which carries only the bound selection and is deliberately `distinctUntilChanged` (an interface appearing *outside* the selection never emits there). The interface ships a default that emits one snapshot, so other platforms keep compiling and behave exactly as today.
- `DesktopNetworkInterfaceService` wires it to the `NetworkStateMonitor` events it already consumes: debounced re-read on each OS network change, plus one immediate read whenever a collector subscribes (so reopening the page re-reads too). Shared on the service scope, which also moves the blocking `NetworkInterface.getNetworkInterfaces()` call off the composition dispatcher.
- `NetworkSettingsContentView` collects that flow instead of doing the one-shot read. The loading row is unchanged: `initial = null` still means "first enumeration in flight".

**Test**

`DesktopNetworkInterfaceServiceSelfHealingTest` gains a case that walks the #4996 shape: with discovery bound to `en0` alone, an ICS-style `192.168.137.1` interface appearing mid-session shows up in the picker (while the bound selection stays `en0`), and re-subscribing re-reads the snapshot with no network event.


guiyanakuang:
Two hardening follow-ups pushed on top:

- **`getAllNetworkInterfaceInfo()` no longer lets a failed enumeration escape.** `NetworkInterface.getNetworkInterfaces()` throws when the platform enumeration itself fails (a Windows IP Helper error, a machine with no interfaces); the per-NIC probes were already wrapped but that top-level call was not — including in the pre-existing `resolveNetworkInterfaceInfos` path, so this guards both call sites at the source. It now logs and reports an empty list, which the resolver already treats as "offline": the persisted preference is left untouched and the next network event re-reads.

  Worth spelling out why this matters more for the new flow than for the old code: `networkRefreshScope` is a `SupervisorJob`, so a throw would not take the scope down — but it *would* kill the `shareIn` coroutine, leaving the settings picker spinning on its loading row for the rest of the process lifetime with no way to recover.

- **`flowOn(ioDispatcher)` on the interface's default implementation**, so a platform that has not wired its own monitor still keeps the blocking enumeration off the collector's dispatcher. The desktop override was already safe (shared on the IO-backed service scope); this closes the same gap for the default path.

No test for the enumeration guard: the failure is a static JDK call, and the existing suite substitutes `getAllNetworkInterfaceInfo()` wholesale, so a test would only assert that `runCatching` catches.

guiyanakuang:
One more commit: the filter in `getAllNetworkInterfaceInfo()` now logs what it drops.

Triaging #4996 ran into the gap: the reporter's log shows the JVM offering exactly one interface, while `ipconfig` shows two — and an interface dropped by `!isUp || isLoopback || isVirtual` was indistinguishable in the log from one the platform never enumerated, because the filter was silent. Skipped interfaces now log name, display name, MAC and the three flags, which is also what lets a log line be matched to an `ipconfig` row.

guiyanakuang:
Reworked the skip logging after reproducing a Windows ICS host locally (Windows 11 + two adapters, ICS enabled, private side on `192.168.137.1/24`) and running a probe against the exact runtime we ship, `jbrsdk-21.0.9-windows-x64-b1163.94`:

```
enumerated 13 interfaces
== name=ethernet_32769 | display=Parallels VirtIO Ethernet Adapter | idx=6
   isUp=true loopback=false virtual=false          addr=/10.211.55.4/24
== name=ethernet_32770 | display=Parallels VirtIO Ethernet Adapter #2 | idx=2
   isUp=true loopback=false virtual=false          addr=/192.168.137.1/24
== name=ethernet_0..5   | display=...-WFP Native MAC Layer LightWeight Filter / QoS Packet Scheduler
   isUp=true loopback=false virtual=false          (no addresses)
== name=ethernet_32768  | Microsoft Kernel Debug Network Adapter     isUp=false
== name=tunnel_32512/13/14 | Teredo / IP-HTTPS / 6to4                isUp=false
```

Two things came out of that, both now in the code:

- A per-interface skip line would have printed **four** lines on every enumeration on any Windows machine — the permanently down tunnel adapters are always there. The skips are now collected into one line per enumeration.
- The pseudo-adapters (`ethernet_0..5`) pass the filter and then contribute nothing, because they carry no IPv4 address — the same silence the filter had. An interface kept but without an IPv4 address is now logged too, so the log accounts for every interface the JVM returned: kept with an address, kept without one, or skipped with the reason.

The probe result itself says the JDK enumerates an ICS adapter correctly in a plain two-adapter setup, so whatever drops it on the reporter's machine in #4996 is not simply "the JVM cannot see ICS adapters". Diagnosing that needs this logging, which is why it lands here.

Issue #5013, opened 2026-09-16
[Windows] Bitwarden is not detected as clipboard source, so it cannot be added to Source Exclusion


maple037:
### How frequently does the bug occur?

Always

### Bug description

On Windows, CrossPaste cannot correctly detect Bitwarden as the source application of copied clipboard content.

Because Bitwarden is not recognized as the clipboard source, it does not appear as an application that can be added to **Source Exclusion**.

As a result, passwords or other sensitive content copied from Bitwarden may still be captured by CrossPaste.

### Steps to reproduce

1. Open Bitwarden on Windows.
2. Copy a password or any text from Bitwarden.
3. Check the copied item in CrossPaste.
4. CrossPaste does not identify the source application as Bitwarden.
5. Try to add Bitwarden to **Source Exclusion**.
6. Bitwarden cannot be selected/added because CrossPaste does not detect it.

### Expected behavior

CrossPaste should recognize copied clipboard content as coming from Bitwarden, so Bitwarden can be added to **Source Exclusion**.

Once excluded, clipboard content copied from Bitwarden should not be recorded or synchronized by CrossPaste.

### Actual behavior

Bitwarden is not recognized as the clipboard source on Windows, so it cannot be added to the Source Exclusion list.

### Environment

- OS: Windows 11
- CrossPaste version: v2.2.0
- Application: Bitwarden Desktop

### Additional information

The same feature works correctly on Linux: CrossPaste can recognize Bitwarden there and Bitwarden can be added to Source Exclusion.

The issue appears to be Windows-specific.

### Stacktrace & log output

```shell

```

### Can you reproduce the bug?

Always

### Version

v2.2.0

### OS

Windows

guiyanakuang:
Thanks for the report. This is now resolved on `main` and will ship in the next release, 2.2.1. It was addressed in three parts:

1. #5015: CrossPaste honors the clipboard hints password managers attach to secrets. On Windows, Bitwarden marks copied passwords with `CanIncludeInClipboardHistory = 0`, so they are no longer recorded or synced, regardless of whether the source app is detected.
2. #5017: the Clipboard sources page now lists only excluded apps, and you can add any app from a picker (running or recent apps, or browse for an `.exe`), so Bitwarden can be excluded even when it is not detected as a source.
3. #5020: manual match rules as a fallback, excluding any source whose name contains a given text.

[...]

PR #5019, opened 2026-09-17
:sparkles: exclude clipboard sources by match rule

### LINKED ISSUE #5018: Clipboard sources: exclude by match rule when the exact source name cannot be picked
## Problem

Source exclusion is an exact string match, and the exact name is not always something the user can pick:

- On Linux a `.desktop` entry without `StartupWMClass` is named by its file stem (e.g. `org.mozilla.firefox`), while the app's windows report a different `WM_CLASS` (e.g. `firefox`). Excluding the entry silently does nothing.
- On macOS a few bundles display a `CFBundleDisplayName`/`CFBundleName` that differs from what `NSRunningApplication.localizedName` reports at runtime.
- Some users simply know part of the name (`bitwarden`) and not the exact string.

## Scope

- A new persisted list `sourceExclusionPatterns` next to the exact exclusions. Older configs get an empty list; no migration.
- A source whose name contains any rule, compared case-insensitively, is not recorded. Substring only; no wildcards or regular expressions.
- The Clipboard sources page gains a "Match rules" section: text input (Enter or button to add, blank rejected, duplicates ignored), one row per rule with a remove button. On Linux the section subtitle points at the `WM_CLASS` case.

Third and last part of #5013.

guiyanakuang:
Closes #5018. Third and last part of #5013. Stacked on #5017; GitHub retargets this PR to `main` once that one merges.

## What

Users can type a text; any clipboard source whose name contains it (case-insensitive) is no longer recorded. This is the fallback for apps whose exact source name cannot be picked: Linux entries whose `WM_CLASS` differs from the desktop file stem, the rare macOS bundle whose display name differs from its runtime name, or simply a user who knows `bitwarden` but not the exact string.

## How

- `DesktopAppConfig.sourceExclusionPatterns`: a JSON string list next to `sourceExclusions`, default empty, so older configs load unchanged.
- `DesktopSourceExclusionService.isExcluded` now returns true for an exact hit or for any rule contained in the source name (`contains(ignoreCase = true)`). `addPattern` trims, rejects blank input and duplicates (ignoring case) and returns whether it added; `removePattern` and `getPatterns` complete the API. The three platform pasteboard services call `isExcluded` as before, so rules apply everywhere without further wiring.
- `SourceExclusionPatternsSection`: header row, an input row (Enter on key-down or the button submits; the button is disabled while blank; the field clears after submit, duplicate or not) and one row per rule with a remove button. Placed between the "Add app" row and the exclusion list. On Linux the subtitle reads "Still recorded after excluding? Add a rule with its WM_CLASS".
- i18n: 4 new keys in all 10 locales, extension translations regenerated.

## Review

Independent strict review: no P0/P1. Adjudicated P2s:

- The Linux hint sat in a single-line, ellipsised subtitle and could be cut off in a narrow panel. Fixed by shortening it in every locale so `WM_CLASS` stays visible.
- Adding a duplicate rule left the input untouched with no feedback. Fixed: the input clears after any submit, since the intent is already met.
- `isExcluded` decodes two small JSON lists per clipboard change instead of one. Accepted as negligible; no caching added.

## Tests

- `DesktopSourceExclusionServiceTest` (5, fast tier, real `DesktopConfigManager` on a temp dir): exact matching is whole-string only, rules match as case-insensitive substrings, rules are trimmed and deduplicated ignoring case, removal resumes recording, exclusions are not added twice.
- `I18nConsistencyTest` covers the new keys.
- Fast tier: 1799 passed, 0 failed. ktlintCheck clean.

## Not verified yet

The rules section's Compose interaction on real machines; tracked with the #5013 verification checklist.


guiyanakuang:
Closed automatically when the base branch of #5017 was deleted. Continued in #5020 against main with the same commits.

